PR #18, opened 2021-09-03
Feat/table sorting

johnflank:


moshloop:
@johnflank The sorting does not persist through refreshes of the grid.

flanksaul:
> @johnflank The sorting does not persist through refreshes of the grid.

Are we still implementing with react-table?

@moshloop Sort could also be valuable to share in the URL.

PR #25, opened 2021-09-09
feat: adds sharable url

flanksaul:


src/components/Canary/state.js
```diff
@@ -0,0 +1,96 @@
...
+
+export function updateFormState(update, url, labels) {
+  const { layout, groupBy, hidePassing, labels: updateLabels = {} } = update;
+  const { labels: decodedLabels = {}, ...rest } = decodeUrlSearchParams(url);
+  const incoming = { ...decodedLabels, ...updateLabels };
+  const newLabels = initialiseLabelState(labels, incoming);
+
+  const groupByDefaults = new Map([
+    ["description", null],
+    ["name", null],
```
moshloop, line 79: Can we make group by name the default ?

src/components/Canary/url.js
```diff
@@ -0,0 +1,24 @@
+const cleaner = (key, value) =>
+  // See re prototype pollution: https://book.hacktricks.xyz/pentesting-web/deserialization/nodejs-proto-prototype-pollution
+  key === "__proto__" || key === "constructor" ? undefined : value;
+
+export function encodeObjectToUrlSearchParams(object) {
```
moshloop, line 5: I think we are over encoding here, the parameters and value should be normal parameters, only the labels/filters  should be encoded as json

src/components/Canary/state.js
flanksaul: Done.

PR #27, opened 2021-09-10
chore: update browserlist

flanksaul:
This is the supported list with the new config (run `npm run checkbrowserslist` to see):

```and_chr 92
and_ff 90
and_qq 10.4
and_uc 12.12
android 92
android 4.4.3-4.4.4
baidu 7.12
chrome 92
chrome 91
chrome 90
chrome 89
chrome 87
chrome 85
edge 92
edge 91
firefox 90
firefox 89
firefox 78
ios_saf 14.5-14.7
ios_saf 14.0-14.4
ios_saf 13.4-13.7
kaios 2.5
op_mob 64
opera 77
opera 76
safari 14.1
safari 14
safari 13.1
samsung 14.0
samsung 13.0```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@flanksource/flanksource-ui",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "private": false,
   "files": [
     "build",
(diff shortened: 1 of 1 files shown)


moshloop:
Why is this necessary ?

flanksaul:
> Why is this necessary ?

Compiling to older JavaScript does two things:

1) it makes the program slower by preventing it from using modern language features in execution – many features since IE 11 are not just syntactic sugar over older features. 
2) it can really bloat bundle size, which increases both load and parse time.

If we want to support older browsers we should do a differential load.

ciju:
We can merge this. Should I rebase it?

PR #30, opened 2021-09-20
Feat/react table rebase

johnflank:


src/components/Canary/table.js
```diff
@@ -1,149 +1,314 @@
...
+    () =>
+      tableData.map((check) => ({
+        ...check,
+        name: GetName(check)
+      })),
+    [tableData]
+  );
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const columns = useMemo(() => Object.values(columnsKeyed), []);
```
flanksaul, line 150: This shouldn't need the lint exclusion?

src/components/Canary/state.js
```diff
@@ -78,7 +80,9 @@ export function updateFormState(update, url, labels) {
     ...Object.entries(newLabels)
   ]);
 
-  const groupByValueOrDefault = groupByDefaults.has(groupBy) ? groupBy : "name";
+  const groupByValueOrDefault = groupByDefaults.has(groupBy)
+    ? groupBy
+    : "no-group";
```
flanksaul, line 85: This appears to have been reverted?

src/components/Canary/table.js
```diff
@@ -1,149 +1,314 @@
...
+  // Hide expander column if there is no grouping
+  useEffect(() => {
+    toggleHideColumn("expander", !hasGrouping);
+  }, [hasGrouping, toggleHideColumn]);
+
+  // Set table's sort state according to url params on page load or url change
+  useEffect(() => {
+    const searchParams = window.location.search;
+    const decodedParams = decodeUrlSearchParams(searchParams);
+    if (decodedParams.sortBy) {
```
flanksaul, line 201: The item (sortBy) by which you wish to sort may no longer exist as the table refreshes. There needs to be a set of defaults and a check to see whether the item exists. If it does not exist, we need to revert to defaults.

src/components/Canary/table.js
```diff
@@ -1,149 +1,314 @@
...
   useEffect(() => {
-    setAggregated(aggregate(title, items));
-  }, [items, title]);
+    const updateURL = (sortBy, sortDesc) => {
+      const searchParams = window.location.search;
+      const decodedParams = decodeUrlSearchParams(searchParams);
+      const newFormState = {
+        ...decodedParams,
+        sortBy,
+        sortDesc
```
flanksaul, line 216: These will override the decoded parameters, which is fine, unless the item doesn't exist.

flanksaul (changes requested):
We need to account for resetting to defaults when the item in the URL doesn't exist as the table is subject to refresh and changing data.

flanksaul (approved):

PR #69, opened 2021-11-15
Feat/settings panel

johnflank:


moshloop:
Can we make it "Auto Refresh" with a default of on? And a dropdown with 10 seconds, 30 seconds, 1 minute, 5 minute, 10 minutes ?

moshloop:
@johnflank the settings need to persist in browser local storage + the  dropdown should be hidden if it is toggled off

johnflank:
Pushed something else to this branch on accident, reverted the changes.

PR #74, opened 2021-11-24
Feat/redesigned modal

johnflank:
branch [feat/settings-panel](https://github.com/flanksource/flanksource-ui/pull/69) should be merged first


moshloop:
Looking good, just a couple things:

* Can we move the history and details into 2 tabs, with history being the default selected tab?
* The font size on the table is a bit small
* The table should scroll, rather than the modal scrolling
* The modal height should be 80% height for >= 1280, otherwise 90%

johnflank:
@moshloop 

> Can we move the history and details into 2 tabs, with history being the default selected tab?

Yes, will work on this alongside the chart

> The font size on the table is a bit small

Can be done as well. I initially reduced the font size because for certain checks the text felt a little suffocated, but I guess slightly increasing the modal width will alleviate that issue.

> The table should scroll, rather than the modal scrolling

OK, I will assign a max-height for the table so that it scrolls if it goes beyond a fixed value. I'll still allow the modal to scroll independently(if it's too tall for the user's screen) since the user's viewport height is something we have no control over.

> The modal height should be 80% height for >= 1280, otherwise 90%

I think this will result in a lot of empty white space at the bottom part of the modal on bigger screens, which might feel awkward. Hard-coding a container's height rather than letting the internal elements decide its height is something I feel we should avoid.

If what we want is a modal that doesn't float/scroll away off-screen, one alternative is to scroll the internal content of the modal. So let me know what you think should be the solution here 👍

johnflank:
This branch is rebased with a new PR: 
https://github.com/flanksource/flanksource-ui/pull/77

Do refer to the above.

PR #77, opened 2021-11-26
Feat/redesigned modal rebased

johnflank:
rebased version of:
https://github.com/flanksource/flanksource-ui/pull/74
https://github.com/flanksource/flanksource-ui/tree/feat/redesigned-modal


moshloop:
At high resolutions the table is scrolling even though the dialog is only taking up 50% of height
Screenshot 2021-11-28 at 20 08 29

johnflank:
@moshloop adjusted the table contents to take up as much height as it needs, if there's space available. Also adjusted so that the Scrollbar only appears within the table, if there's content overflow

image

moshloop:
@johnflank  If the status and details tabs are different heights than the modal height changes, can we make it fixed?

johnflank:
> @johnflank If the status and details tabs are different heights than the modal height changes, can we make it fixed?

Just fixed that jarring height-change issue. I implemented a JS solution that detects which tab has the highest content height, then uses that value as the "fixed" content height.

Doing this we also avoid the "lots of empty space" issue from the modal potentially being stretched vertically in big screens.

PR #119, opened 2022-01-23
V2

moshloop:


src/components/HypothesisBuilder/components/HypothesisNode/index.js
```diff
@@ -0,0 +1,316 @@
+import React, { useEffect, useState } from "react";
```
JIamep20, line 1: I've been on projects where all components were placed in index.js files. Over time, it becomes difficult to search for the desired file among hundreds of identical ones, and in ide, the same tab names are also not encouraging. Therefore, in our company it is customary to name the file in the same way as the main component that is exported from it is called. On the other hand, different conventions in one project are also not good. Let me know if we need to change our code to match yours in this regard.

src/components/Layout/sidebar.js
```diff
@@ -0,0 +1,248 @@
...
+function classNames(...classes) {
+  return classes.filter(Boolean).join(" ");
+}
+
+// eslint-disable-next-line import/no-default-export
+export default function SidebarLayout({ navigation, userNavigation }) {
+  const [sidebarOpen, setSidebarOpen] = useState(false);
+  const [user, setUser] = useState(null);
+  if (user == null) {
+    getUser().then((user) => {
```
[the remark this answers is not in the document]
moshloop: react-query looks nice :) please open an issue for our backlog to implement, it probably makes sense todo react-table at the same time

src/components/HypothesisBuilder/components/HypothesisNode/index.js
moshloop: I think we can use that convention and then use index.js for exporting the actual classes which may be in different folders/files

PR #121, opened 2022-01-25
fix: card selection on click and menu items / button effects

JIamep20:
1. hover effects for dropdown menu button
2. hover effects for dropdown menu items
3. card selection on card mouse click / enter, space press

### DIFF
--- a/src/components/HealthSummary/summary.js
+++ b/src/components/HealthSummary/summary.js
@@ -6,16 +6,39 @@ import { Chip } from "../Chip";
 function getChipsFromSummary(summary) {
   const chips = [];
   if (summary.healthy > 0) {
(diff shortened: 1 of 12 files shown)


moshloop:
@JIamep20 I don't think we want a new hover effect for the button - should be more like the bell in top right and avatar menu
for card selection click on card is good - but no need for space selection and the outline that comes with it

JIamep20:
@moshloop 
> but no need for space selection and the outline that comes with it

I did it because 2 eslint rules required at least one key press event handler and role alongside onClick, i decided that empty stubs are not acceptable here. So do you want to leave only enter selection, on remove keyboard control at all?

>  don't think we want a new hover effect for the button - should be more like the bell in top right and avatar menu

in other words, would you like us to repeat the styles from the header?

PR #135, opened 2022-01-31
refactor: merge small and medium card into one

JIamep20:
merge small and medium card into one, rework styles to classes despite the fact that it is not recommended according to tailwindcss documentation


moshloop:
Can we rebase this?

* Small has MetricsFooter instead of MetricsHeader
* M, L, XL - are different max-width and heights on the card itself
* Metrics Header in M has the label above the property and and in L+ to the right of

PR #138, opened 2022-02-01
Feat/incidents list

JIamep20:


src/components/Incidents/IncidentList/index.js
```diff
@@ -1,29 +1,146 @@
...
+    >
+      <td
+        colSpan={2}
+        className="px-6 py-4 text-darker-black col-span-2 text-sm leading-5 font-medium"
+      >
+        {title}
+      </td>
+      <td className="px-3 py-3">
+        <div className="flex flex-row items-center">
+          <Icon name={severityInfo.icon} />
```
JIamep20, line 109: i decided this is more universal. heroicons may not contain all the required icons

moshloop: We use https://react-icons.github.io which has pretty much all free icon sets - The included icons are intended primarily for product logos.

JIamep20: @moshloop done. Anything else here ?

src/components/Incidents/IncidentList/index.js
```diff
@@ -1,29 +1,145 @@
-import { Link } from "react-router-dom";
+import React, { useMemo } from "react";
+import { useNavigate } from "react-router-dom";
+import cx from "clsx";
```
moshloop, line 3: why the rename from clsx to cx ?

moshloop:
merged into v2 branch already

PR #143, opened 2022-02-04
Feat/react select dropdown

JIamep20:
1. installed react-select
2. implemented "Select all" feature
3. Wrapped it into "react-hook-form" controller
4. Styled a little to follow design in figma


src/components/Select/Select.js
```diff
@@ -0,0 +1,196 @@
...
+  singleValue: (styles) => ({
+    ...styles,
+    fontSize: "16px",
+    lineHeight: "24px"
+  })
+};
+
+export const Select = ({ name, control, ...props }) => {
+  return (
+    <Controller
```
moshloop, line 150: The `control` is optional you can specify either control or onSelect

moshloop:
Also there is this border: 

Screenshot 2022-02-04 at 13 33 42

src/components/Select/Select.js
JIamep20: What's the point in additional onSelect props? You can use original onChange prop... Or you want it to transform somehow value.

moshloop: either onSelect or onChange is fine - by control must be optional

JIamep20: done, removed border in the input

moshloop:
@JIamep20  please can you resolve conflicts and squash your commits ?

PR #144, opened 2022-02-08
Feat/log filter tweaks

johnflank:
Changed the order of the log filters


moshloop:
So the id is not the topologyId - it the id associated with the type e.g. pod id, this comes from the external_id on the topology

Issue #141, opened 2022-02-03
Add back style for small and large cards


moshloop:

DmitriiMelnikOnix:
#156

PR #156, opened 2022-03-07
Feat: styles for small and large cards

DmitriiMelnikOnix:
#141


moshloop:
The layout does not look right - the metrics are always horizontal. never vertical

DmitriiMelnikOnix:
> The layout does not look right - the metrics are always horizontal. never vertical

It seems that the `grid-cols-3` class was not applied in the preview version. It is very strange. I replaced `grid` to `flex`

moshloop:
@DmitriiMelnikOnix looks better now  but the topology selector modal still is not showing the checkboxes for selection

moshloop:
@DmitriiMelnikOnix the checkboxes are fixed, but when clicking add the dialog does not close

moshloop:
@OleksandrSekker please can you sign the CLA ?

Issue #122, opened 2022-01-25
Modal Improvements


moshloop:
Create 3 sizes as attributes:

- [ ] Small
- [ ] Medium
- [ ] Full (Full-screen height - gap based on the screen size full-screen width - gap to a max of 1024
- [ ] Header and footer sections must be fixed on the modal body scrolls

i.e size should replace cardClass, contentClass, cardStyle

moshloop:
i.e. 

```javascript
  <Modal
      open={isOpen}
      onClose={handleModalClose}
      cardClass="w-full"
      contentClass="h-full pt-10 pl-10 pb-10 pr-7 font-inter"
      cardStyle={{ maxWidth: "77.6rem" }}
      closeButtonStyle={{ padding: "2.375rem 2.375rem 0 0" }}
      containerClass="p-4"
      hideActions
    >
      <h1 className={cx("text-2xl font-semibold mb mb-6", titleClassName)}>
        {title}
      </h1>
      <div
        className="grid gap-4 w-full px-0.5 overflow-x-auto"
        style={{ gridTemplateColumns: "repeat(5, minmax(12.375rem, 1fr))" }}
      >
        {topologies.map((topology, index) => (
          // eslint-disable-next-line react/no-array-index-key
          <div key={index}>
            {topology.map((item) => (
              <TopologyCard
                size="small"
                topology={item}
                key={item.id}
                selectionMode
                selected={checked.indexOf(item.id) > -1}
                onSelectionChange={(state) => {
                  toggleChecked(item.id, state);
                }}
              />
            ))}
          </div>
        ))}
      </div>
      <div
        className={cx("flex justify-end mt-7 align-baseline", footerClassName)}
      >
        {!hideCounter && (
          <p className={cx("flex items-center", footerTextClassName)}>
            <span>{`${checked.length} cards selected`}</span>
          </p>
        )}
        <button
          type="button"
          className={cx(
            "py-3 px-6 bg-dark-blue rounded-6px text-white ml-6 hover:bg-warm-blue",
            submitButtonClassName
          )}
          onClick={() => onSubmit(checked)}
          disabled={checked.length === 0}
        >
          {submitButtonTitle}
        </button>
      </div>
    </Modal>
```

```javascript
  <Modal
      open={isOpen}
      onClose={handleModalClose}
      size="medium"
      hideActions
title={title}
actions={[action1]}
    >
  
      <div
        className="grid gap-4 w-full px-0.5 overflow-x-auto"
[...]

PR #157, opened 2022-03-07
Fix/modal improvements

DmitriiMelnikOnix:
#122


moshloop (changes requested):
Please create an example demonstrating these 2:

*  Full (Full-screen height - gap based on the screen size full-screen width - gap to a max of 1024
* Header and footer sections must be fixed on the modal body scrolls

moshloop:
Please check the style the padding around the buttons in the footer looks wrong, there is also a ghost sidebar, and the full size needs to stretch to the full height of the page, only the width is limited to 1024

Can you also add an example with lots of text to see how the scroll works ?

DmitriiMelnikOnix:
@moshloop I can't find the ghost sidebar. Can you take a screenshot?

moshloop:
@DmitriiMelnikOnix 
Screenshot 2022-03-11 at 08 48 16
Screenshot 2022-03-11 at 08 47 39


The gap between the bottom of the dialog and the window is also missing

DmitriiMelnikOnix:
> The gap between the bottom of the dialog and the window is also missing

The gap is exist if close netlify menu. I don't think this menu needs to be taken into account. It is only in deploy-preview

moshloop:
@VincentWild69 please can you sign the CLA ?

Issue #71, opened 2021-11-16
JUnit Viewer


moshloop:

DmitriiMelnikOnix:
@moshloop do you have any more JSON examples with errors or other cases?

PR #174, opened 2022-03-21
feat: JUnit viewer

DmitriiMelnikOnix:
#71


moshloop:
@DmitriiMelnikOnix Can we put this into a sortable tree-table with columns for description, descriptions, and Summary (passed/failed) ?

DmitriiMelnikOnix:
@moshloop do you have any more JSON examples with failed or other cases?

moshloop:
@DmitriiMelnikOnix it is the same date e.g.

`Untitled suite in /Users/niko/Sites/casperjs/tests/suites/casper/agent.js
(2/3)
0.256ms`

`2` is passed, `3` is total ( so not passed/total, keep as is passed/total)

DmitriiMelnikOnix:
@moshloop can I use the `Table` component from `src/components/Canary/table.js` or should I create a new tree-table component?

moshloop:
@DmitriiMelnikOnix I think create a new tree table component use react-table expansion or another tree-component if that doesn't work

DmitriiMelnikOnix:
@moshloop do you have any style preference for the table? Is there some example of which I can take the styles for the table?

moshloop:
@DmitriiMelnikOnix you can use https://docs.microsoft.com/en-us/azure/devops/pipelines/test/review-continuous-test-results-after-build?view=azure-devops as a reference

moshloop:
Also the hierarchy is based on suite.name ->  (group by test.classname) -> test.name

moshloop:
@DmitriiMelnikOnix  looking better, however:

1) In this screenshot the last row should be indented under `tests/suites/casper/agent`
Screenshot 2022-03-30 at 07 59 04

2) Missing counts per level
3) Missing duration column

moshloop:
@DmitriiMelnikOnix  looking better - Can we format the duration? 
Also the count should be passed of total, not failed of total, 
The counts should also rollup to the top so a parent count = sum(children)

moshloop:
@DmitriiMelnikOnix - looking good, can we add a filter to hide passing tests (with a param to to the combonent that defaults to hidePassing) and another parameter to hide the header with pass/failed etc ?

moshloop:
@DmitriiMelnikOnix  the failed count in this screenshot should be 1 not 0

Screenshot 2022-04-17 at 10 25 05

moshloop:
@DmitriiMelnikOnix The hide header option should not hide the status column, it should hide the header with Failed / Passed / Total count status etc..

DmitriiMelnikOnix:
> @DmitriiMelnikOnix the failed count in this screenshot should be 1 not 0

It was a data error. Initially all tests were passed and I added failed test to check the table

Issue #168, opened 2022-03-20
Hypothesis Modal - Title is in body


moshloop:
The editable title should be passed through to the Modal as the title so that is rendered correctly

PR #175, opened 2022-03-21
fix: move hypothesis title to modal title

DmitriiMelnikOnix:
#168


src/components/HypothesisBuilder/components/HypothesisDetailsTitle/index.js
```diff
@@ -0,0 +1,63 @@
+import React, { useEffect, useRef } from "react";
```
moshloop, line 1: please rename file to `HypothesisBuilder/hypothesis-title.js`

moshloop (changes requested):

Issue #171, opened 2022-03-20
Comment component refactoring


moshloop:
* extract a reusable comment component from hypothesis-builder
e.g.

```javascript
<Comment data={data}  markup="@[__display__](user:__id__)" onChange=... />
```

i.e. mentionsStyle and renderSuggestion should be private to the component

PR #176, opened 2022-03-23
Fix/comment component refactoring

DmitriiMelnikOnix:
https://github.com/flanksource/flanksource-ui/issues/171

### DIFF
--- a/src/components/HypothesisBuilder/comments.js
+++ b/src/components/HypothesisBuilder/comments.js
@@ -45,23 +45,12 @@ export function CommentsSection({
     console.log("type tag", type, "value", id);
   };
 
-  const commentData = useMemo(
-    () => [
-      {
-        data: users,
-        markup: "@[__display__](user:__id__)",
-        trigger: "@"
-      }
-    ],
-    [users]
-  );
-
   return (
     <div className={rest.className} {...rest}>
       {titlePrepend}
       <div>
         <CommentInput
-          data={commentData}
+          data={users}
           value={commentTextValue}
           onChange={setCommentTextValue}
         />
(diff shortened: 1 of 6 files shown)


src/components/HypothesisBuilder/comments.js
```diff
@@ -69,69 +32,39 @@ export function CommentsSection({
...
-      },
-      [otherText[0]]
-    );
-  };
+  const commentData = useMemo(
+    () => [
+      {
+        data: users,
+        markup: "@[__display__](user:__id__)",
+        trigger: "@"
```
moshloop, line 53: Can we move markup and trigger as attributes on Comment Input and then default them ?

DmitriiMelnikOnix: We can if we use only mentions of users. But if you need to make tags mentions and user mentions then the current example will be better

moshloop: Even if we add support for tags, it would be in addition `#` to users not instead of

DmitriiMelnikOnix: Do you suggest doing the following?
```
<CommentInput data={data} markup="@[__display__](user:__id__)" trigger="@" ... />
```
But how then to make support a few triggers and markup?
In this case, we can use either @ or # trigger. But not both at the same time

moshloop: I suggest that interface but default the markup and trigger, at a later point in the future if we decide to handle tags we can come up with a solution then

DmitriiMelnikOnix: Ok I updated interface

Issue #155, opened 2022-03-04
Topology Dropdown


moshloop:
* [ ] Create a select component that does a remote search on topology name
* [ ] Show icon, title, and status 

```jsx
<TopologyDropdown multiple=true parent= />
```

PR #191, opened 2022-04-04
Feat: topology dropdown

DmitriiMelnikOnix:
https://github.com/flanksource/flanksource-ui/issues/155


moshloop:
@DmitriiMelnikOnix 

- the healthy / unhealthy should only display in the dropdown, not in the selected list
- The min chars to query should be removed, if I click the dropdown I want to see all topologies, and then I can filter down

Issue #78, opened 2021-11-28
Date/Time Range picker


moshloop:
Similar to the grafana range picker 
<img width="546" alt="Screenshot 2021-11-28 at 19 38 41" src="https://user-images.githubusercontent.com/1489660/143779373-8d88bd0e-9472-498d-84b8-828236cd5ef3.png">

moshloop:
The time range component should then emit `onChange(from,to)` with from, to fields for initial values

PR #214, opened 2022-04-18
Time range picker

DmitriiMelnikOnix:
https://github.com/flanksource/flanksource-ui/issues/78


moshloop:
@VincentWild69 please sign CLA

This is looking good - can we style it with the tailwind classes and theme ?

VincentWild69:
@[moshloop](https://github.com/moshloop) For the calendar, which is used in the component, stylization is provided using classes prepared in advance in it. Therefore, the use of tailwind in this case is not appropriate. I also encountered a bug: some classes from tilewind do not work (for example: w-fit, some color schemes), and a different way of styling is needed to apply these properties. There are also quite a lot of interactive elements, and with the classic CSS file, working with them looks clearer. I used tailwind to the maximum, but in this case I think it is appropriate to leave some style in the css file, according to the arguments above.

VincentWild69:
@moshloop Can we leave the styling as it is? ^

moshloop:
@VincentWild69  I don't particularly care how the styling is done, but it must match the look and feel of the existing components If I put a dropdown and a timepicker next to each other they should look the same

moshloop:
@VincentWild69  please rebase and add storybooks

Prashanth-Nelli:
Handled this requirement in PR [78](https://github.com/flanksource/flanksource-ui/issues/78)

PR #218, opened 2022-04-19
replace back button with link - 211

### LINKED ISSUE #211: Config Viewer - Breadcrumbs
Replace the "back" button with a link in the top nav bar back to home

gyanpgupta:
Closes #211

### DIFF
--- a/src/pages/config/config-details.js
+++ b/src/pages/config/config-details.js
@@ -1,6 +1,6 @@
 import { useEffect, useState } from "react";
 import toast from "react-hot-toast";
-import { useNavigate, useParams } from "react-router-dom";
+import { useNavigate, useParams, Link } from "react-router-dom";
 import {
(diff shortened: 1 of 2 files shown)


src/pages/config/config-details.js
```diff
@@ -91,16 +91,13 @@ export function ConfigDetailsPage() {
           ? `Config Details for ${configDetails.name}`
           : "Config Details"
       }
+      button={<Link className="underline mr-3"
+        to={`/topology`}>
```
moshloop, line 95: @gyanpgupta why the change from /config to /topology ?

gyanpgupta: @moshloop , before it was redirecting to previous page but as it is mentioned in the task to remove back button & add a link which redirect to home page and '/topology' is the route for the homepage

PR #241, opened 2022-04-28
Feat/add back totals

johnflank:
Add totals/filtered statistics to Canary sidebar that was previously overlooked when porting canary from v1 to v2.

2nd Filtered card will disappear if `total checks` is equal to `filtered checks`

image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -27,7 +27,6 @@ import { RsDemoPage } from "./pages/Examples/rs-demo";
 import { TopologyPage as ExamplesTopologyPage } from "./pages/Examples/Topology/topology-page";
 import { TopologySelectorModalPage } from "./pages/Examples/TopologySelectorModalPage/TopologySelectorModalPage";
 import { HealthPage } from "./pages/health";
-import { Canary } from "./components/Canary";
 
 const queryClient = new QueryClient({
   defaultOptions: {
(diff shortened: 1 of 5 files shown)


src/components/Canary/index.js
```diff
@@ -98,6 +104,39 @@ export function Canary({
   return (
     <div className="flex flex-row place-content-center">
       <SidebarSticky topHeight={topLayoutOffset}>
+        <div className="mb-4">
+          <CheckStat
+            containerClass="shadow-md rounded-lg p-5 mb-2"
+            valueContainerClass="items-end"
```
moshloop, line 110: Can we move these into Checkstat as defaults ?

src/components/Canary/index.js
```diff
@@ -98,6 +104,39 @@ export function Canary({
...
+            value={checks.length || 0}
+            append={
+              <div className="ml-2 text-sm text-gray-600">total checks</div>
+            }
+          />
+          {filteredCount !== checks.length && (
+            <CheckStat
+              containerClass="shadow-md rounded-lg p-5 mb-2"
+              valueContainerClass="items-end"
+              title="Check Filtering"
```
moshloop, line 121: Can we match the wording that we used previously ?

src/App.js
```diff
@@ -127,6 +129,10 @@ export function App() {
               element={<TypologyDropdownDemo />}
             />
           </Route>
+          <Route
```
moshloop, line 132: Can we remove this example and move it to a storybook ?

moshloop (changes requested):

src/App.js
johnflank: Sure thing, will refer to the existing Storybook files we have

moshloop:
@johnflank please resolve conflicts

johnflank (commented):
Ready to merge

PR #243, opened 2022-05-02
Table view explorer

JordanAtanasov12:
Implemented a table-like, file explorer component:
 - Formatted and beautified date and size strings
 - Added Search functionality
 - Added Paging functionality
 - Added file reader, displaying functionality
 - Given log sample
 - Had great fun, implementing it :)

https://user-images.githubusercontent.com/42078209/166262327-31bebfb6-abc7-490e-9c8f-913fa20e99c4.mp4


src/components/FileExplorerTable/tableFileView.js
```diff
@@ -0,0 +1,154 @@
...
+    //Render Table component with Pagination
+    return (
+        <>
+            <input
+                className="search-input"
+                value={filterInput}
+                onChange={handleFilterChange}
+                placeholder={"Search by file name"}
+            />
+            <table {...getTableProps()}>
```
johnflank, line 52: Can the table be styled to resemble the other tables we already have? For instance, take a look at the canary table (components/canary/table.js) at `health` page or the `ConfigListTable` component in `config-list` page

src/index.css
```diff
@@ -170,4 +170,136 @@
     width: 3.125rem;
     height: 3.125rem;
   }
+
```
johnflank, line 173: If these styles are specifically only meant for a particular component (in this case, the file explorer table), we should 
put the stylesheet in the component's folder (and only imported by that particular component) to segregate these classes away from everything else.

For an instance of this, look at `/index.js` and `/index.css` in `src/components/Modal`

src/pages/file-explorer-page.js
```diff
@@ -0,0 +1,107 @@
...
+    if (numberString > 1024) {
+      numberString = numberString / 1024;
+      return numberString + "MB";
+    }
+    else {
+      return "0." + numberString + "MB";
+    }
+  }
+
+  function formatDate(dateString) {
```
johnflank, line 33: Will there be any sort of `dateString` edge cases that might cause issues here? Unless we're absolutely sure the `dateString` will **always** be in the format of YYYYMMDD this type of checking can be dangerous. Just making sure.

PR #256, opened 2022-05-24
fix: logs page infinite load & health check api infinite load issue

Prashanth-Nelli:


src/components/Canary/index.js
```diff
@@ -101,24 +105,33 @@ export function Canary({
...
         setIsLoading(false);
       });
   }, 1000);
 
-  useMemo(() => {
+  useEffect(() => {
     handleFetch();
-    const interval = setInterval(handleFetch, refreshInterval);
-    return () => clearInterval(interval);
+    const intervalRef = setInterval(handleFetch, refreshInterval);
```
ciju, line 117: Would it be better to use `setTimeout` instead of `setInterval`? Timeout would have implicit back-pressure. With interval, there would be request every 15 sec. With timeout, next request will only happen 15s after the last one. And throttle might not be needed. 

Not a blocker. If no issues with current code, we can merge it.

ciju:
LGTM.

PR #261, opened 2022-05-27
handled logs loading on query changes

Prashanth-Nelli:
- handled logs loading on query changes
- removed current behaviour of loading logs on page load.

### DIFF
--- a/src/components/Logs/Table/logs-table.js
+++ b/src/components/Logs/Table/logs-table.js
@@ -130,7 +130,11 @@ export const LogsTable = ({ logs, actions, variant }) => {
             return (
               <tr key={key} {...restHeaderGroupProps}>
                 {headerGroup.headers.map((column) => (
-                  <th key={column.id} {...column.getHeaderProps()}>
+                  <th
+                    key={column.id}
+                    {...column.getHeaderProps()}
+                    className="py-2"
+                  >
                     {column.render("Header")}
                   </th>
                 ))}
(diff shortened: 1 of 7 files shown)


src/components/Logs/Table/logs-table.js
```diff
@@ -127,7 +127,11 @@ export const LogsTable = ({ logs, actions, variant }) => {
             return (
               <tr key={key} {...restHeaderGroupProps}>
                 {headerGroup.headers.map((column) => (
-                  <th key={column.id} {...column.getHeaderProps()}>
+                  <th
+                    key={column.id}
+                    {...column.getHeaderProps()}
+                    className="pt-2 pb-2"
```
ciju, line 133: `py-2`?

src/pages/logs-page.js
```diff
@@ -44,15 +44,15 @@ export const logTypes = [
 ];
 
 export function LogsPage() {
-  const [logsIsLoading, setLogsIsLoading] = useState(true);
+  const [logsIsLoading, setLogsIsLoading] = useState();
```
ciju, line 47: Could be explicit about the loading states. E.g. `initial | loading | done`

src/pages/logs-page.js
```diff
@@ -99,16 +99,17 @@ export function LogsPage() {
...
   }, []);
 
   useEffect(() => {
-    loadLogs();
-  }, [loadLogs, topologyId]);
+    const subscription = watch(() => {
+      loadLogs();
+    });
+    return () => subscription.unsubscribe();
+  }, [watch]);
```
ciju, line 112: We don't need to watch for `topologyId` here? Including `loadLogs` in should also be ok?

Prashanth-Nelli: we are watching all controls changes here so topologyId is also covered. loadlogs not required as it's reference won't change.

ciju (approved):

PR #273, opened 2022-05-31
feat: Show tags in Config List page

ciju:
Closes: https://github.com/flanksource/flanksource-ui/issues/272 
Closes: https://github.com/flanksource/flanksource-ui/issues/210


moshloop:
Can we also add a very small vertical gap between tags ?

PR #280, opened 2022-06-02
fixed logs page bugs & add useLoader hook

Prashanth-Nelli:


src/hooks/useLoader.ts
```diff
@@ -0,0 +1,28 @@
...
+  function computeLoadingState(showLoader: boolean) {
+    setIdle(false);
+    setRequestsCounts((val) => {
+      val = showLoader ? val + 1 : val - 1;
+      val = val < 0 ? 0 : val;
+      return val;
+    });
+  }
+  return [idle, loading, loaded, computeLoadingState];
+}
```
ciju, line 28: Do we need tracking of multiple loading states? Logic would be simpler to reason if we are only tracking one. 

The complexity of multiple async ops is anyhow something we need to tackle at call site with promise cancellation etc. Keeping the useLoader straightforward would make it easier to reason. E.g. all it does is `idle → loading ↔ loaded`. No counting.

Also, instead of multiple variables, states could be represented in one variable. `state: ['loading',  'loaded', 'idle']`.

And `useLoader.ts` could export helper functions to check the state. call site e.g.

```js
import { useLoader, isLoading, isLoaded } from '../hooks/useLoader';

const { loadingState, setLoading } = useLoader();

const asyncFn = () {
  setLoading(true);
}

return (
<div>
  {isLoading(loadingState) && <Loading ... />}
  {isLoaded(loadingState) && <Data ... />}
</div>
)
```

Single state would be easier to understand and debug.

src/constants/index.js
```diff
@@ -1,5 +1,2 @@
-export const LoadingStates = {
-  idle: "idle",
-  loading: "loading",
-  loaded: "loaded"
-};
+export * from "./hypothesis-statuses";
+export * from "./incident-priority";
```
ciju, line 2: Blanket imports and exports are usually not a good idea. Clashes. Difficulty in reading code (where is the symbol imported from) etc.

ciju (commented):
No blocking Issues. Put some comments for consideration.

ciju (approved):

src/hooks/useLoader.ts
```diff
@@ -0,0 +1,28 @@
...
+    setLoading(requestsCount !== 0);
+    setLoaded(requestsCount === 0 && !idle);
+  }, [requestsCount]);
+
+  function computeLoadingState(showLoader: boolean) {
+    setIdle(false);
+    setRequestsCounts((val) => {
+      val = showLoader ? val + 1 : val - 1;
+      val = val < 0 ? 0 : val;
+      return val;
```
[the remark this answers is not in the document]
Prashanth-Nelli: I feel this should be `val => Math.max(0, showLoader ? val + 1 : val - 1)`

tailwind.config.js
```diff
@@ -134,7 +134,8 @@ module.exports = {
         "btn-round-sm": "2.125rem",
         "btn-round-md": "2.375rem",
         "btn-round-lg": "2.625rem",
-        "btn-round-xl": "3.125rem"
+        "btn-round-xl": "3.125rem",
+        "144": "36rem"
```
ciju, line 138: What is this for?

PR #281, opened 2022-06-02
fix: change health page default grouping to `canary_name`

ciju:
Closes https://github.com/flanksource/flanksource-ui/issues/278


moshloop:
I think Group Name should map to canary_name, and still be the default  - There are some groups with 5+ individual checks

ciju:
Fixed. Although, I don't understand all the places the wrong variable (`config_name`) was used. Renamed all occurrences with `config_name`. Code has too many indirections. The column name shows up as `config_name`. It's picked up from url. Again, would require some effort to fix.

moshloop:
This was due to some API level changes - Could we remove the name in brackets in the table header  ?

PR #284, opened 2022-06-06
fix: fixed payload structure and moved pod/node selector to the header

Prashanth-Nelli:


src/components/Loading/index.js
```diff
@@ -1,8 +1,8 @@
 import { Oval } from "react-loading-icons";
 
-export function Loading({ text = "Loading..." }) {
+export function Loading({ text = "Loading...", ...props }) {
   return (
-    <div className="flex justify-center items-center">
+    <div className={"flex justify-center items-center " + props.className}>
```
ciju, line 5: Could use `clsx("flex justify-center items-center", props.className)`. otherwise if no className in props, it will show up as ` undefined` class.

src/pages/logs-page.js
```diff
@@ -212,15 +195,17 @@ export function LogsPage() {
...
-        <div>
-          <h1 className="text-xl font-semibold">
-            Logs
-            {topology && (
-              <span className="text-gray-600">
-                / {topology.name || topology.text}
-              </span>
-            )}
-          </h1>
+        <div className="flex">
```
ciju, line 198: is this div needed? if `flex` styling is needed, could it be part of SearchableDropdown's className?

ciju (approved):

PR #290, opened 2022-06-08
Fix incident creation flows from Topology/Logs/Config pages along with some cleanup

ciju:
Closes: https://github.com/flanksource/flanksource-ui/issues/294
Closes: https://github.com/flanksource/flanksource-ui/issues/293


moshloop:
The only issue I could find is that on the initial redirect to the incident the topology card is not shown, only after refresh

ciju:
Ya, an existing bug. Will fix it in future commits around incidents page.

PR #291, opened 2022-06-08
fix: topology page cleanup & breadcrumbs issue fixes

Prashanth-Nelli:


moshloop:
So the red / green blocks should also be clickable as they filter the selection down

moshloop:
Also the card heights should all be the same on the same row, different rows can have different heights but not in the same row. That height should also have a max of 5 lines of text, anything over that should scroll just that section

src/App.js
```diff
@@ -95,7 +95,14 @@ export function IncidentManagerRoutes({ sidebar }) {
       <Route path="" element={<Navigate to="/topology" />} />
 
       <Route path="topology" element={sidebar}>
-        <Route path=":id" element={<TopologyPage url="/canary/api" />} />
+        <Route path=":clusterId" element={<TopologyPage url="/canary/api" />}>
+          <Route path=":groupId" element={<TopologyPage url="/canary/api" />}>
+            <Route
+              path=":podOrNodeId"
```
moshloop, line 101: We only have pod and nodes at the moment but the topology components could be anything

src/components/Topology/topology-breadcrumbs.js
```diff
@@ -1,62 +1,118 @@
...
-      <Link
-        to={`/topology/${_topology.id}`}
-        className="flex flex-nowrap hover:text-gray-500 my-auto "
-      >
-        <Icon name={_topology.icon} size="xl" className="mr-1" />
-        {_topology.name || _topology.title}
+      <Link to="/topology" className="hover:text-gray-500 ">
+        Topology
       </Link>
+      {cluster && (
```
moshloop, line 80: The breadcrumbs can't be hardcoded like this, as the hierarchy is arbritrary

src/pages/Examples/Topology/topology-view.js
```diff
@@ -9,7 +9,7 @@ export const TopologyView = ({
   setSelectionMode,
   toggleChecked
 }) => (
-  <div className="font-inter flex leading-1.21rel">
+  <div className="flex leading-1.21rel">
```
moshloop, line 12: Can we move the example to a storybook ?

src/components/Topology/topology-breadcrumbs.js
Prashanth-Nelli: I went with approach because of following reason.
- Currently we are not returning proper parent_id when we fetch topology by id
   `curl 'http://localhost:3004/canary/api/topology?id=018138c7-c631-ecd2-6b0c-cf6559ebebc2' \
  -H 'Accept: application/json' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Cookie: _ga=GA1.1.27620775.1611673700; amplitude_id_eb9e4621db21faf713ee50c637f3cf36=eyJkZXZpY2VJZCI6IjcxMzgyYzZiLTYyMjEtNDMxNi1iY2RiLTRkMWQ0OGQyOGY2ZFIiLCJ1c2VySWQiOm51bGwsIm9wdE91dCI6ZmFsc2UsInNlc3Npb25JZCI6MTYxMjAyNTc0MTE4MCwibGFzdEV2ZW50VGltZSI6MTYxMjAyNTg2MzI0MiwiZXZlbnRJZCI6MCwiaWRlbnRpZnlJZCI6MCwic2VxdWVuY2VOdW1iZXIiOjB9; _ga_M4NQSZJKMF=GS1.1.1613889114.8.1.1613891920.0; amplitude_id_7f4b058e1a8c4a8746473d8f92d1497f=eyJkZXZpY2VJZCI6ImEzZmVhNTVmLTdjNTMtNGM1Yy04N2RjLTQxOWU5MjYxMDFkY1IiLCJ1c2VySWQiOiIyMTgiLCJvcHRPdXQiOmZhbHNlLCJzZXNzaW9uSWQiOjE2MzI5MjM0OTgzNjIsImxhc3RFdmVudFRpbWUiOjE2MzI5MjQzNDkwMjAsImV2ZW50SWQiOjMzMywiaWRlbnRpZnlJZCI6NDQ5LCJzZXF1ZW5jZU51bWJlciI6NzgyfQ==; _hjid=e312216c-e783-456f-a540-c29c0546adbb; _hjSessionUser_2207626=eyJpZCI6IjRlOTliMDM3LTE1Y2EtNTZjNC1iMDE3LWE2NmI4YjQ2OWM2YSIsImNyZWF0ZWQiOjE2NDMzODc3OTA5NTUsImV4aXN0aW5nIjp0cnVlfQ==; __gads=ID=b7153cc4ae6e4c0b-22bd4e49e1d00070:T=1646676290:RT=1646676290:S=ALNI_MYQhzsuypq1E4Ni7mB1sVrdHFdwVQ; _ga_94LF6Z0YYR=GS1.1.1650807302.204.0.1650807302.0' \
  -H 'DNT: 1' \
  -H 'Pragma: no-cache' \
  -H 'Prefer: return=representation' \
  -H 'Referer: http://localhost:3004/topology/018138c7-c627-98b8-65e2-c9fd9fb6e9db/018138c7-c631-ecd2-6b0c-cf6559ebebc2' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36' \
  -H 'sec-ch-ua: " Not A;Brand";v="99", "Chromium";v="102", "Google Chrome";v="102"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  --compressed`

Is there any limit on topology depth  ?

src/components/Topology/topology-card.js
```diff
@@ -124,20 +139,22 @@ export function TopologyCard({
...
-                  className={
-                    index === _topology.properties.length - 1
-                      ? "mb-0"
-                      : "mb-2.5"
-                  }
-                />
-              ))}
-            </div>
-            <div className="w-med-card-right pl-1 py-4 pr-5 overflow-y-auto">
+            {Boolean(properties.length) && (
```
ciju, line 142: I think this would show `false` in html, if `properties.length` is 0.

Prashanth-Nelli:
> So the red / green blocks should also be clickable as they filter the selection down

I have added filtering which does the same in the search toolbar

src/constants/index.js
```diff
@@ -1,2 +1,14 @@
...
+  node: "Node",
+  created: "Created At",
+  "instance-type": "Instance Type",
+  zone: "Zone",
+  os: "OS",
+  containerRuntime: "Container Runtime",
+  kernel: "Kernel",
+  kubeProxy: "Kubernetes proxy",
+  kubelet: "Kubelet "
+};
```
ciju, line 14: If this is only used at one place, might be better to keep in the file where it's used. Otherwise, while reading code, one would have to jump files to figure what the map is.
UTY.

src/components/Topology/topology-card.js
```diff
@@ -124,20 +139,22 @@ export function TopologyCard({
...
+                {properties.map((property, index) => (
+                  <Property
+                    key={property.name}
+                    property={property}
+                    className={
+                      index === _topology.properties.length - 1
+                        ? "mb-0"
+                        : "mb-2.5"
+                    }
+                  />
```
ciju, line 153: Unrelated, but could `space-y-{n}` simplify the `className` here? UTY.

ciju (commented):
- UTY is **Up To You**. Not blocking for merge.
- Have you considered keeping breadcrumbs by pushing the name for the previous path in a component which is rendered throughout the lifetime of Topology page? The Breadcrumb component was built with that in mind. That way, you don't have to hard code for cluster, node etc. Again, not a blocker. If current UI suffices, we can go ahead with this approach.

src/components/Topology/topology-card.js
Prashanth-Nelli: JSX, won't render boolean values, they should be converted to strings to get them rendered. I think we are good here.

src/App.js
```diff
@@ -95,7 +95,14 @@ export function IncidentManagerRoutes({ sidebar }) {
...
+              element={<TopologyPage url="/canary/api" />}
+            ></Route>
+          </Route>
```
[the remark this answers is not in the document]
Prashanth-Nelli: Moshe said depth is not fixed, it can vary topology topology, I need to revisit breadcrumb generation logic.

src/pages/Examples/Topology/topology-view.js
Prashanth-Nelli: I'll do it in the follow up PR

PR #295, opened 2022-06-09
Feat/query builder changes

Prashanth-Nelli:


moshloop:
* So the query (`/query` endpoint)  is instead of Type/Text/Search (which uses the `/db/config_item` endpoint)  - we need a toggle between them
* I don't see any button to click to run the query
* The preset selection takes up too much space - Ideally we can fit this into the same row in the header

Prashanth-Nelli:
> * So the query (`/query` endpoint)  is instead of Type/Text/Search (which uses the `/db/config_item` endpoint)  - we need a toggle between them
> * I don't see any button to click to run the query
> * The preset selection takes up too much space - Ideally we can fit this into the same row in the header

currently we are only re fetching configs when one of the saved query selected. I'll add a way to run a query without saving it

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,175 @@
+import { useEffect, useState } from "react";
+import { TrashIcon } from "@heroicons/react/solid";
```
ciju, line 2: We can use react-icons here? react-icon seems to have most icons. Long term, we can try removing heroicons.
UTY

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,175 @@
...
+    },
+    {
+      null: {
+        id: null,
+        name: "None",
+        description: "None",
+        value: null
+      }
+    }
+  );
```
ciju, line 94: Map version might be easier to read. Something like:

```js
Object.fromEntries(queryList.map(({id, description}) => [id, {
        id,
        name: id,
        description,
        value: id
}]
.concat([null, {name: "None", description: "None"])));
```

UTY

Prashanth-Nelli:
> **A reviewer left a comment:**
> 
> screenshot
> 
> Browser metadata
> [Open Deploy Preview](https://deploy-preview-295--goofy-euclid-75956c.netlify.app/config?query=) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=62a2bda71dcafa0009042433&commentID=62a2d7f1ac4cbc65f9d49698&resolution=resolved)

Done

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,298 @@
...
+      </div>
+      <Transition
+        as={Fragment}
+        enter="transition ease-out duration-100"
+        enterFrom="transform opacity-0 scale-95"
+        enterTo="transform opacity-100 scale-100"
+        leave="transition ease-in duration-75"
+        leaveFrom="transform opacity-100 scale-100"
+        leaveTo="transform opacity-0 scale-95"
+      >
```
ciju, line 272: Is this significant enough to merit the verbosity? UTY

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,298 @@
...
+        </div>
+      </Modal>
+    </div>
+  );
+};
+
+function QueryBuilderActionMenu({ onOptionClick, options }) {
+  return (
+    <Menu as="div" className="relative inline-block text-left">
+      <div>
```
ciju, line 259: `div` wrapper needed? UTY.

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,298 @@
...
+        id: queryItem.id,
+        label: queryItem.description,
+        type: "action_saved_query",
+        context: {
+          ...queryItem
+        }
+      });
+    });
+    return data;
+  }, [queryList, selectedQuery]);
```
ciju, line 65: I think the logic would be simpler if actions are rendered explicitly and their actions handled explicitly. Instead of having everything in array. Also, from UI perspective, it makes sense to separate action (save, save as, update) from saved queries (even if it's just a horizontal separator).

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,298 @@
...
+      handleRunQuery();
+    }
+  }, []);
+
+  const handleSearch = (value) => {
+    setQuery(value);
+    setParams({
+      query: value
+    });
+  };
```
ciju, line 91: With query directly from params, this could be written as
```js
const handleSearch = (query) =>  setParams({ query });
```

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,298 @@
...
+    if (!queryToRun) {
+      toastError("Please provide a query before running");
+      return;
+    }
+    refreshConfigs([]);
+    try {
+      const result = (await getConfigsByQuery(queryToRun))?.data?.results || [];
+      result.forEach((item) => {
+        item.tags = JSON.parse(item.tags);
+      });
```
ciju, line 123: Better to use immutable operations. Easy to mess with mutation. Also, probably better handled inside `getConfigsByQuery`, including the data.results check logic?

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,298 @@
...
+  const handleSaveQuery = () => {
+    setSavedQueryValue("");
+    if (!query) {
+      return toastError("Please provide query details");
+    }
+    setModalIsOpen(true);
+  };
+
+  const handleRunQuery = async () => {
+    const queryToRun = query;
```
ciju, line 113: Any particular reason for this?

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,298 @@
...
+                  )}
+                </Menu.Item>
+              );
+            })}
+          </div>
+        </Menu.Items>
+      </Transition>
+    </Menu>
+  );
+}
```
ciju, line 298: Does it make sense to have separate component for query actions? Haven't tried coding so not sure how it would look.

src/pages/logs-page.js
```diff
@@ -90,7 +90,9 @@ export function LogsPage() {
   const [topologyId, setTopologyId] = useState(searchParams.get("topologyId"));
   const [externalId, setExternalId] = useState(searchParams.get("externalId"));
   const [type, setType] = useState(searchParams.get("type"));
-  const [start, setStart] = useState(timeRanges[0].value);
+  const [start, setStart] = useState(
+    searchParams.get("start") || timeRanges[0].value
+  );
```
ciju, line 95: Get start directly from params, instead of setting another state?

ciju (commented):
Not for this PR, but try using typescript. I am finding it quite useful catching issues and code completion etc.

src/components/QueryBuilder/index.js
```diff
@@ -0,0 +1,298 @@
...
+  updateSavedQuery
+} from "../../api/services/configs";
+import { Modal } from "../Modal";
+import { toastError, toastSuccess } from "../Toast/toast";
+import { useLoader } from "../../hooks";
+import { TextInputClearable } from "../TextInputClearable";
+
+export const QueryBuilder = ({ refreshConfigs, className, ...props }) => {
+  const [params, setParams] = useSearchParams();
+  const [query, setQuery] = useState(params.get("query") || "");
```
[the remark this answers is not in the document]
Prashanth-Nelli: Sure

src/pages/logs-page.js
ciju: Didn't get that. Would there be an issue with something like:
```diff
modified   src/pages/logs-page.js
@@ -90,9 +90,8 @@ export function LogsPage() {
   const [topologyId, setTopologyId] = useState(searchParams.get("topologyId"));
   const [externalId, setExternalId] = useState(searchParams.get("externalId"));
   const [type, setType] = useState(searchParams.get("type"));
-  const [start, setStart] = useState(
-    searchParams.get("start") || timeRanges[0].value
-  );
+
+  const start = searchParams.get("start") || timeRanges[0].value;
 
   const [topologies, setTopologies] = useState([]);
   const [logs, setLogs] = useState([]);
@@ -223,7 +222,7 @@ export function LogsPage() {
             name="start"
             className="w-40 mr-2 flex-shrink-0"
             items={timeRanges}
-            onChange={(e) => setStart(e)}
+            onChange={(e) => setSearchParams({ start: e })}
             value={start}
           />
         </>
```

Prashanth-Nelli: @ciju I have url params syncing logic at one place. This code is already in place so followed the same pattern here. May I know why this approach is better compared to previous one.

src/components/QueryBuilder/index.js
ciju: https://github.com/flanksource/flanksource-ui/commit/4824a1ea2442e3e7e41c98189108f70760882ca6 is this more readable?

src/pages/logs-page.js
ciju: Ignore. Will setup a pattern to follow (and reasoning), later. You can close resolve this issues.

src/components/QueryBuilder/index.js
Prashanth-Nelli: @ciju Thanks for the suggestion, I am already doing the refactoring.

moshloop:
* Can we add a separator in the dropdown and menu header "Saved Queries" just below ?
* Loading a saved query doesn't update the URL, if you load and then edit the query it then saves

Prashanth-Nelli:
> * Can we add a separator in the dropdown and menu header "Saved Queries" just below ?
> * Loading a saved query doesn't update the URL, if you load and then edit the query it then saves

Made changes as suggested.

src/components/TextWithDivider/index.css
```diff
@@ -0,0 +1,20 @@
...
+  flex-grow: 1;
+}
+
+.text-divider::before {
+  margin-right: 5px;
+}
+
+.text-divider::after {
+  margin-left: 5px;
+}
```
ciju, line 20: why not use tailwind?

Prashanth-Nelli: I am not aware of the classes which work similar to ::before & ::after classes. If I find any alternative I will work on removing this stuff in follow up PR. 

Can we you please create a doc around the coding guidelines and link it in the README.md file, that will give us an idea what is allowed and what is not.

moshloop:
* @Prashanth-Nelli  when I save a new query, the URL is removed and the input field blanks out
* Loading a query, adds it to the input field and URL, but doesn't actually run it

Prashanth-Nelli:
> * The delete query should probably also move to the dropdown

done

Prashanth-Nelli:
> * @Prashanth-Nelli  when I save a new query, the URL is removed and the input field blanks out
> * Loading a query, adds it to the input field and URL, but doesn't actually run it

done

PR #298, opened 2022-06-10
fix: fixed logs loading on duration change

Prashanth-Nelli:


ciju (approved):
LGTM. Although, another option is to use searchParam for custom duration. That way it could be shared also.

ciju:
General advice is to have single source of truth. And if the param is in url, url should be the source of truth, because user can copy paste the url.

Issue #179, opened 2022-03-29
Update hypothesis title


moshloop:
[hypothesisStatus Icon] [Title]                                                                       [avatar1][avatar2] X

For the avatars we might also need to create a new AvatarGroup that overlays them like this: https://tailwindui.com/components/application-ui/elements/avatars#component-e4a92edf4f12acee331db4190019b4ef 

When clicking the statusIcon it drops down to show the icons + text

PR #300, opened 2022-06-14
Incident pages changes

ciju:
- Add delete option in hypothesis bar (doesn't confirm the action).
- Show avatar group on incident list and hypothesis bar. Closes: https://github.com/flanksource/flanksource-ui/issues/179
- Add stories for HypothesisBar, HypothesisTitle, HypothesisNode


moshloop:
It's looking much better: :)

* Can we remove the white background on the close button
* Can we switch to a large or xlarge dialog size 
* Why is the first user avatar + name, but the 2nd avatar only? 
* Maybe the state should be an icon only on the left of the title, and then clicking on it drops down the menu? 
* I think the comments should be descending 
* Should the comments not use the same avatar as everything else ?
* I thought the checkboxes in the logs were to delete specific lines, we should probably remove and switch to the compact view 
* For the hyptothesis deletion, we probably want to add popup that says: `Are you sure you want to delete the hypothesis or would you rather mark this hypothesis as Disproven`  With a Disprove and Delete buttions
* On new hypothesis create the title should be in header, and the input focused on launch

ciju:
> * Why is the first user avatar + name, but the 2nd avatar only? 
First is the user who created the story. 
2nd is a group avatar. Could be many users with overlapping avatar icon. Could do conditional, show name if only one etc. But we could do that later, if needed?

Will create separate issue for rest of the comments. Should I prioritize these over the ConfigItem story?

ciju:
Issue for tracking the comments: https://github.com/flanksource/flanksource-ui/issues/307

PR #312, opened 2022-06-17
Changes around HypothesisBuilder

ciju:


moshloop:
@ciju  looking good:

* On the delete popup the button is focused on open - I don't think anything should be focused
* The x is also missing
* The buttons don't have a gap between them

src/api/services/comments.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const getCommentsByHypothesis = (hypothesisId: string) =>
+  resolve<Comment[]>(
+    IncidentCommander.get(
+      `/comment?select=*,created_by(id,name,avatar)&hypothesis_id=eq.${hypothesisId}&order=created_at.desc`
+    )
+  );
+
+export const createComment = async ({
```
Prashanth-Nelli, line 29: can we have type info here.

ciju:
```js
}: NewComment) =>
  resolve<Comment>(
```
those are the types. Did you mean something else?

src/components/Comment/CommentText.tsx
```diff
@@ -0,0 +1,37 @@
...
+        return (
+          <React.Fragment key={idx}>
+            {otherText[idx + 1] || ""}
+            <button
+              type="button"
+              key={tagId}
+              onClick={() => onClickTag("user", tagId)}
+              className="bg-blue-200 rounded"
+            >
+              {tagDisplay || ""}
```
Prashanth-Nelli, line 30: can we move this logic to the top where we are calculating tagDisplay value

src/components/Hypothesis/CreateHypothesis/index.tsx
```diff
@@ -66,22 +62,28 @@ export const CreateHypothesis = ({ node, api, onHypothesisCreated }) => {
           status: getValues("hypothesis.status")
         }
       });
-      // const newNode = newNodeResponse.data[0];
-      // if (isEmpty(getValues("comment.text"))) {
-      //   await handleComment(newNode.id, getValues("comment.text"));
-      // }
+      console.log({ newNodeResponse });
```
Prashanth-Nelli, line 65: can we remove the log statements.

src/components/Hypothesis/HypothesisTitle/index.tsx
```diff
@@ -43,22 +47,27 @@ export const HypothesisTitle = ({ node, api }: Props) => {
...
-          {node?.created_by?.name}
-        </p>
-      </div>
-      {Boolean(involved.length) && (
-        <AvatarGroup maxCount={5} users={involved} size="sm" />
+      {Boolean(involved.length || node?.created_by) && (
+        <span className="pl-4">
+          <AvatarGroup
+            maxCount={5}
+            users={[node?.created_by].concat(involved)}
```
Prashanth-Nelli, line 67: What happens when **node?.created_by** is undefined, do we filter out all the undefined entries ?

src/components/Comment/CommentText.tsx
ciju: Curious, why is this is a concern? Moving it to top makes line longer and makes it part of ignore lint.

src/components/Hypothesis/CreateHypothesis/index.tsx
```diff
@@ -66,22 +62,28 @@ export const CreateHypothesis = ({ node, api, onHypothesisCreated }) => {
...
-        <div className="text-sm font-medium text-gray-700 mb-1.5">
-          <label htmlFor="name">Name</label>
-        </div>
+    <Modal
+      open={isOpen}
+      onClose={onHypothesisCreated}
+      wrapWith={({ children }) => (
+        <form
+          onSubmit={(...args) =>
+            console.log(...args) || handleSubmit(onSubmit)(...args)
```
Prashanth-Nelli, line 79: I think we need to remove the log statement here, WDYT.

src/components/Comment/CommentText.tsx
Prashanth-Nelli: As we are already do the value computation at the top, why do we need to handle value absence case in the jsx expression we can do that during the computation itself.

ciju: A readability stylistic preference? It doesn't add confusion, and at least to make, makes it easier to read. Otherwise I have to parse a long line to understand the behavior.

Prashanth-Nelli: I generally prefer doing value calculation at once, UTY.

PR #313, opened 2022-06-20
feat: add time range picker component

Prashanth-Nelli:
I have handled below review feedback on the PR

- [x] Make calendar occupy full width
- [x] Improve section headers styling
- [x] Remove range selection on calendar & replace it with individual selection
- [x] Improve date picker input styling
- [x] Update relative list so that It will support all the values similar to Grafana time range picker.
- [x] Make range picker value holder width to be content adjustable.
- [x] Improve submit button styling & modify the label to Done/Submit.
- [x] Focus to/from input value holder when it's counterpart value is provided.
- [x] Auto close calendar on value selection
- [x] Allow manual edit & validate manually edited values 
- [x] Make the font size on absolute vs relative time ranges consistent.
- [x] We should be displaying the same formatting in the from/to fields as we do in the label i.e. no seconds or millis and no time zone conversion
- [x] The Recently used ranges should not wrap


src/components/TimeRangePicker/RecentlyRanges.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+RecentlyRangesFC.propTypes = {
+  recentRanges: PropTypes.arrayOf(PropTypes.objectOf(PropTypes.string)),
+  applyTimeRange: PropTypes.func
+};
+
+RecentlyRangesFC.defaultProps = {
+  recentRanges: [],
+  applyTimeRange: () => {}
+};
```
ciju, line 45: Defaults can be set directly in the function param. Easier to read. Eg:
```js
({
  recentRanges = [],
  applyTimeRange = () => {}
}: RecentlyRangesProps) => (
```
Typescript takes care of validations. Same applies to other components in the PR.

src/components/TimeRangePicker/RecentlyRanges.tsx
```diff
@@ -0,0 +1,47 @@
...
+        </button>
+      ))}
+    </div>
+  </div>
+);
+
+RecentlyRangesFC.propTypes = {
+  recentRanges: PropTypes.arrayOf(PropTypes.objectOf(PropTypes.string)),
+  applyTimeRange: PropTypes.func
+};
```
ciju, line 40: Unnecessary with typescript. Same applies to other components in the PR.

src/components/TimeRangePicker/TimePickerCalendar.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+    </div>
+    <div>
+      <Calendar
+        locale="en"
+        selectRange
+        next2Label={null}
+        prev2Label={null}
+        nextLabel={<FaAngleRight size="18px" color="#272727" />}
+        prevLabel={<FaAngleLeft size="18px" color="#272727" />}
```
ciju, line 38: Should we use one of the tailwind colors? setting `className="text-color..." should be enough.

src/components/TimeRangePicker/TimeRangeList.tsx
```diff
@@ -0,0 +1,89 @@
...
+            <label htmlFor={id} className="cursor-pointer py-1.5 px-2">
+              {option.display}
+            </label>
+            <input
+              type="checkbox"
+              className="opacity-0 cursor-pointer"
+              checked={isChecked(option, currentRange)}
+              onChange={() => {}}
+              name="range-checkbox"
+              id={id}
```
ciju, line 64: Would a simpler `id`, like option index suffice? Second param in map callback is index.

src/components/TimeRangePicker/RecentlyRanges.tsx
Prashanth-Nelli: Let me check.

moshloop:
When selecting a relative time range I don't see anything change ?

I am expecting it to be:

From: 15 minutes
To: now

Can we also get variations with a) nothing  pre-selected, b) an absolutate range and c) a relative range

src/components/TimeRangePicker/RecentlyRanges.tsx
Prashanth-Nelli: Let me know, if I am wrong.

ciju: Typescript makes it irrelevant. If types check out. Runtime should not be an issue.

Prashanth-Nelli:
> When selecting a relative time range I don't see anything change ?
> 
> I am expecting it to be:
> 
> From: 15 minutes To: now
> 
> Can we also get variations with a) nothing pre-selected, b) an absolutate range and c) a relative range

- added storybook variants
- earlier range values are converted to date, now modified it to use ranges documented in the Grafana time range controls.
- previous storybook doesn't handle change event of TimeRangePicker because of which range selection is not populated properly, I have fixed it now.

moshloop:
* The width is pretty large - it is replacing a 170px input now, We can also remove the inline label as it is not necassary
* On the date pickers - the range is very unintuitive - I expect clicking on from to popup a window to click on a date, once clicked it closes, and moves the focus to, I click on the icon select a date and it closes again. 
* In the popup, the width of the date fields can be increased it should take up the the full width
* Can we style the button more like the search bars? i.e. inside the input ?
*

moshloop:
Screenshot 2022-06-26 at 11 09 23

* The font size on absolute vs relative time ranges is not consistent. 
* We should be displaying the same formatting in the from/to fields as we do in the label i.e. no seconds or millis and no time zone conversion
* The Recently used ranges should not wrap

Issue #304, opened 2022-06-15
CreateOrLinkIncidentDialog: Allow creation of new Incident,Hypothesis,... 


moshloop:
- [ ] Replace CreateIncidentDialog and LinkIncidentDialog with CreateOrLinkIncidentDialog. 
- [ ] Allow for search and if not creation, from the dropdown, for incident and hypothesis

Issue #269, opened 2022-05-30
Improve styling of Create Incident page/dialog


moshloop:
From topology/config/logs/ etc..

PR #321, opened 2022-06-24
Feat/link existing incidents

### LINKED ISSUE #270: Create Incident - Link to incident/hypothesis from
Select existing hypothesis or create new.

Current flow allows for creating incidents from Topology/Logs/Config pages. Add option to search and assign the selected evidence (selected topology/component, lines of logs or config lines) to an existing hypothesis (of an incident).

ciju:
Closes #270, #269, #304


moshloop:
Not sure the button groups work - They are not configurable now but the priority, status, and type will be in the future  - You could have priorities like "Severity 1 - All transactional business systems down" or status of "Pending - Waiting on vendor to respond".

ciju:
Should we consider the option when we face the issue? Anyhow we would have to make changes to support this (e.g. having different priorities than current one). And the effort for changing this to dropdown should not be much. And for the current options, the UX seems congested, with all the dropdowns. Let me know if you still prefer to have the dropdown now.

PR #234, opened 2022-04-22
Feat/hypothesis comment view 

DmitriiMelnikOnix:
https://github.com/flanksource/flanksource-ui/issues/187

PR #333, opened 2022-07-04
[Incident Details] UX changes for comments and incidents

ciju:
Closes: #306

Closes: #234

- [x] Can we move the comment button inline into the field, and then make it enter to add, like a chat window ?
- [x] Move the "Comment" title as a placeholder in the input box
- [x] Remove the "no comments" placeholder
- [x] Make the time fixed e.g. `10:43am`, `9:00am (yesterday)`, `10:00pm, Tuesday 21st June 2022`


moshloop:
* Can we move the comment button inline into the field, and then make it enter to add, like a chat window ?
* Move the "Comment" title as a placeholder in the input box
* Remove the "no comments" placeholder
* Make the time fixed e.g. `10:43am`, `9:00am (yesterday)`, `10:00pm, Tuesday 21st June 2022`

PR #339, opened 2022-07-12
337 testing bug fixes

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli looking good - The topology cards need to be a fixed width in the topology page - otherwise it becomes very difficult to scan over them. Maybe when used as evidence they can shrink - but not grow

moshloop:
The cards are still too wide:
Screenshot 2022-07-15 at 08 58 34

moshloop:
* The show more, isn't really working well (I think maybe just small ellipsses on its own text line in grey, and scrolls bar on hover)
* the horizontal scroll should never show as we are ellipsing the text
* Can we update the storybook for the cards to include more data (i.e. cards that overflow and cards that don't) 
* I think we also need to size the cards based on a) whether the sidebar is open and b) the browser - i.e the cards should fit on the row without any wasted space on the right

ciju:
Instead of having separate stories for each type of responder, would it make sense to show all responders in one story? That way there would be less code, files and boilerplate.

Prashanth-Nelli:
> Instead of having separate stories for each type of responder, would it make sense to show all responders in one story? That way there would be less code, files and boilerplate.

   I don't see any problem here to be frank, if we worry about writing too much boilerplate code then we should ideally be writing storybooks at page level. That will honestly decrease the amount of boiler plate code we write for each of the components.

moshloop:
Can we save the width in local storage? and then also leave out the exact size - and change the title "Card Width" and reduce the size, and right align it

moshloop:
Can you make it look more like this: 
Screenshot 2022-07-19 at 17 43 42

Issue #328, opened 2022-06-29
Hypothesis details - Indicate with multiple evidences of the same type with a badge


moshloop:
<img width="821" alt="Screenshot 2022-06-29 at 10 25 21" src="https://user-images.githubusercontent.com/1489660/176377133-22d83dd5-c3d5-41f1-8d9c-3847166841fa.png">

PR #348, opened 2022-07-20
UI updates to hypothesis bar

ciju:
- [x] Show count for multiple evidence of same type
- [x] Move delete, create solution actions to menu
- [x] Improve styling for side line to child hypothesis
- [x] Move edit action to menu
- [x] Can we switch back to the input with the tick or x to complete editing? Can we also make the input have some padding so it doesn't take up the full height of the bar
- [x] The hypothesis should expand/collapse on clicking on the title

Closes: https://github.com/flanksource/flanksource-ui/issues/328


moshloop:
- [ ] Can we switch back to the input with the tick or x to complete editing? Can we also make the input have some padding so it doesn't take up the full height of the bar
- [ ] The hypothesis should expand/collapse on clicking on the title

ciju:
Pushed the changes.

moshloop:
* It looks like the click target is not the bar so if you click on the top or bottom of the bar, it doesn't do anything
* Can we expand the entire tree of children ?

ciju:
Expanding all children is becoming tricky. Right now, it works if the page has `Show Comments` enabled. When disabled, click only expands current hypothesis node. If children need to be expanded in this case also, I can spend some more time on it.

moshloop:
I don't think we should have a Show Comments, rather just a Expand / Collapse All

ciju:
Updated with expand/collapse rewording.

ciju:
Collapse expand behavior should be fixed now.

PR #349, opened 2022-07-20
feat: add config item search component

Prashanth-Nelli:


moshloop:
For some reason the story book is not working, 

Can we add linked config items e.g.

```jsx
<ConfigItem  name="instance" type="EC2Instance">
<ConfigItem name="ebs"  items="$.block_device_mappings", name="$.DeviceName" value="$.Ebs.VolumeId"/>
</ConfigItem>
```

i.e. First let the user select an EC2Instance, and then show the EBS Volumes from that config ?

.storybook/middleware.js
```diff
@@ -0,0 +1,10 @@
+const { createProxyMiddleware  } = require("http-proxy-middleware");
+module.exports = function expressMiddleware(router) {
+  router.use(
+    "/config",
+    createProxyMiddleware({
+      target: "https://incident-commander.canary.lab.flanksource.com",
+      changeOrigin: true
+    })
+  );
+};
```
ciju, line 10: Let's not use external server for story book tests. You could mock request response for individual story.

src/components/ConfigItem/ConfigItem.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <ConfigItem
+        className="w-96"
+        type={type}
+        value={selectedItem}
+        autoFetch={true}
+        onSelect={(item) => {
+          setSelectedItem(item);
+          setDependentSelectedItem(null);
+        }}
+        dependentConfigItems={[
```
moshloop, line 37: Shouldn't this be a child rather than a property?

src/components/ConfigItem/ConfigItem.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+};
+
+const ConfigItemDropDown = ({ type }: { type: string }) => {
+  const [selectedItem, setSelectedItem] = useState(null);
+  const [dependentSelectedItem, setDependentSelectedItem] = useState(null);
+  return (
+    <div className="flex flex-col">
+      <ConfigItem
+        className="w-96"
```
moshloop, line 29: I think this should be defaulted - we may also need to add support for labels, and descriptions for the fields. otherwise it might be hard to fit this into a form. 

The dependants should also still render when the parent is null, but in a disabled state - it makes the ui more discoverable, and less prone to dialog resizing etc.

src/components/ConfigItem/ConfigItem.stories.tsx
Prashanth-Nelli: We can do that, I just went with this approach to do solve it first. Adopting this to children way of loading is straightforward. I am working on it.

src/components/ConfigItem/ConfigItem.stories.tsx
Prashanth-Nelli: I am working on moving this to child way of loading that will solve and provide support for all the use cases mentioned above.

Prashanth-Nelli:
> @Prashanth-Nelli can we add in other form fields - i.e. replace Project in Jira responder form with a config dropdown ?

Do we have Jira specific configs in our DB ?.

moshloop:
Can we add the issue types in as well ?

moshloop (changes requested):

src/components/ConfigItem/ConfigItem.stories.tsx
```diff
@@ -0,0 +1,206 @@
...
+          }}
+          isDisabled={!selectedItem}
+        >
+          <ConfigItem
+            name="issueType"
+            className="w-96 mt-1"
+            type="JIRA"
+            value={selectedIssue}
+            autoFetch={false}
+            itemsPath="$..task.status[*]"
```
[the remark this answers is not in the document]
Prashanth-Nelli: Yes, that is how it works. parent ConfigItems will feed the selected item as data to child Config Item components and json paths will be applied on the selected value.

src/components/IncidentDetails/ResponderTypes/Jira/index.tsx
```diff
@@ -1,58 +1,115 @@
...
+            <ConfigItem
+              type="Jira"
+              control={control}
+              name="issueType"
+              value={issueType}
+              autoFetch={false}
+              onSelect={(selected) => {
+                setIssueType(selected);
+              }}
+              itemsPath="$..task.status[*]"
```
moshloop, line 93: This is wrong - it selects all status from all projects, it needs to show only the issue types for the selected project,

Issue #356, opened 2022-07-22
Fix topology unroll


moshloop:
https://incident-commander.canary.lab.flanksource.com/topology/01822507-2c29-4037-b2d9-57fd6560ef81

The components unrolled and duplicated +2 requests are made 1 with depth=0 and 1 without - it should only be without

PR #357, opened 2022-07-25
Remove topology response unroll

ciju:
Closes: #356

Depends On: https://github.com/flanksource/canary-checker/pull/571


moshloop:
@ciju - I am getting `Response has multiple components for the id` ?

ciju:
There are some issue with the response from backend. Tarun is looking into one part of it, other will discuss on standup.

ciju:
Issue should be fixed now.

PR #362, opened 2022-07-26
feat: types derived from spec

ciju:


moshloop:
I like this idea - but maybe we should publish the openapi spec on the incident-commander repos, and then download the spec when generating the type definitions ?

ciju:
ya, final version should do that. and manual triggered update of spec on the repo (for diff tracking etc). for now, the focus of the MR is to explore how types would be extracted out of spec and used in the app. The MR right now has type extraction part. Working on the integration of types to the api request/response.

Issue #343, opened 2022-07-17
Update sidebar design


moshloop:
https://tailwindui.com/components/application-ui/application-shells/multi-column#component-6e22fa5715141e2f842a9c887bcfb7a3

moshloop:
![Uploading Screenshot 2022-07-19 at 17.44.41.png…]()

Issue #317, opened 2022-06-21
CRUD - Health Check / Canary


moshloop:
canaries table in canary-checker

Issue #316, opened 2022-06-21
CRUD - Team


moshloop:
- [ ] Create a Settings tab in the navbar, with a Teams sub item
- [ ] List, Create, Update, Delete Teams
- [ ] For the [spec](https://github.com/flanksource/incident-commander/blob/e1e52156f1a4a9540ef1f24a4eff38ea1ceac814/db/migrations/003_teams.sql#L9) field, provide a JSON editor

moshloop:
- [ ] Create a Settings tab in the navbar, with a Teams sub item
- [ ] List, Create, Update, Delete Teams
- [ ] For the [spec](https://github.com/flanksource/incident-commander/blob/e1e52156f1a4a9540ef1f24a4eff38ea1ceac814/db/migrations/003_teams.sql#L9) field, provide a JSON editor

Issue #315, opened 2022-06-21
CRUD - Component


moshloop:
components table in canary-checker

PR #367, opened 2022-07-29
CRUD - Teams, Config Scraper, System template, Component, Health Check / Canary

ciju:
- [x] Create a Settings tab in the navbar, and sub item for each of the table
- [x] List, Create, Update, Delete table
- [x] For the respective specs field, provide a JSON editor

- Teams: `teams` table [spec](https://github.com/flanksource/incident-commander/blob/e1e52156f1a4a9540ef1f24a4eff38ea1ceac814/db/migrations/003_teams.sql#L9)
- Config Scraper: `config_scraper` table in `config-db`
- System template: `system_template` table in `canary-checker`
- Component: `components` table in `canary-checker`
- Health check: `canaries` table in `canary-checker`

All them would have similar schema:

```sql
CREATE TABLE table_name (
  id UUID DEFAULT generate_ulid() PRIMARY KEY,
  name TEXT NOT NULL,
  spec JSONB null,
  source TEXT NULL, -- The CRD source of the rule, if specified the rule cannot be edited via API
  created_at timestamp NOT NULL DEFAULT now(),
  updated_at timestamp NOT NULL DEFAULT now()
)
```

Closes: #316
Closes: #318
Closes: #314
Closes: #315
Closes: #317
Closes: #343


package.json
```diff
@@ -8,6 +8,7 @@
     "scripts/serve-build.js"
   ],
   "dependencies": {
+    "@codemirror/lang-json": "^6.0.0",
```
moshloop, line 11: I think we would want to use monaco instead - We then get full openapi schema support, intellisense for javascript etc...

PR #374, opened 2022-08-10
fix: stop background health refresh api call's on health page unmount

Prashanth-Nelli:


ciju (approved):
lgtm. haven't tested :)

PR #377, opened 2022-08-16
feat: add responder switch to team selection

Prashanth-Nelli:
- [x] Clicking on add responder will show a popup with list of teams.
- [x] User selects any one of the team and after selection he will be shown with add responder flow based on the responders[0] team spec.
- [x] Added Teams details should be shown on the incident details page right side similar to responders list.
- [x] responder list will be: icon {team Name} external_id displayed in the second line


moshloop:
@Prashanth-Nelli  can we put the responders on the single line with the id in () without an "External ID" label
If there is only single responder, also skip the responder selection

Prashanth-Nelli:
> @Prashanth-Nelli can we put the responders on the single line with the id in () without an "External ID" label If there is only single responder, also skip the responder selection

Done

PR #371, opened 2022-08-09
Pending tasks on schema resource pages

ciju:
- [x] Sidebar scrolls with some space on bottom
- [x] Cancel should close create schema page dialog
- [x] Change delete to red with trash icon and confirmation
- [x] Make dialog footer like the Add responder dialog
- [x] Update navbar icons
- [x] Icon dropdown selector for team icon - only for teams table
- [x] Vertical form
- [x] Move create hypothesis from bar menu to inline (After comments)
- [x] Remove comments and evidence section from create hypothesis form

- [x] Can we make the title full width of the dialog ?
- [x] When adding a new 2nd hypothesis the old title is pre-filled
- [x] Change icon to AiOutlinePlusCircle
- [x] Move add solution/factor to below comments

Closes: #376

PR #378, opened 2022-08-16
Inline create incident

ciju:
Remove dialog and inline incident bar with focused title, for new incident


ciju:
Closed in favor of: https://github.com/flanksource/flanksource-ui/pull/371

PR #392, opened 2022-08-23
feat: add owner & topology type filters

Prashanth-Nelli:


moshloop:
Screenshot 2022-08-25 at 13 27 55
* Still seems to be an alignment issue - Maybe we should switch to react-select for all the dropdowns? And then also use https://react-select.com/advanced#experimental popout mode by default ?
* Can we change the cog to FaCog

Prashanth-Nelli:
> https://react-select.com/advanced#experimental

I'll fix the cog icon part. Other item can be handled as a new issue.

PR #394, opened 2022-08-25
Move to NextJS

ciju:


moshloop:
I am getting a method not allowed when selecting a component to view logs on

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+ENV NEXT_TELEMETRY_DISABLED 1
+
+ENV NEXT_PUBLIC_APP_DEPLOYMENT=${APP_DEPLOYMENT}
+RUN NEXT_STANDALONE_DEPLOYMENT=true npm run build
+
+# Production image, copy all the files and run next
+FROM node:16-alpine AS runner
+WORKDIR /app
+
+ENV ORY_KRATOS_URL=${ORY_KRATOS_URL}
```
moshloop, line 39: I don't think this can be part of the dockerfile

chart/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "latest"
+
+imagePullSecrets: []
+nameOverride: "flanksource-ui"
+fullnameOverride: "flanksource-ui"
+
+# Environment for deployment
+env:
+  - name: ORY_KRATOS_URL
```
moshloop, line 19: We can't default this

next.config.js
```diff
@@ -0,0 +1,44 @@
...
+    // your project has type errors.
+    // !! WARN !!
+    ignoreBuildErrors: true
+  },
+  async rewrites() {
+    return [
+      {
+        source: "/api/configs_db/:path*",
+        destination:
+          "https://incident-commander.canary.lab.flanksource.com/config/db/:path*"
```
moshloop, line 14: Do these not need to come from the environment variables ?

Dockerfile
ciju: Didn't get that. It's dynamic environment variable. Needed for the application to figure out the kratos backend.

ciju: It can be set at the runtime. Or picks up value from the time it was built.

next.config.js
ciju: Ya, they can. Just porting the existing `_redirects` netlify config.

chart/values.yaml
ciju: ack, could you add this as secret to the to the aws-sandbox. Will update the chart with the URL from secret there.

next.config.js
```diff
@@ -0,0 +1,44 @@
...
+        destination:
+          "https://incident-commander.canary.lab.flanksource.com/db/:path*"
+      },
+      {
+        source: "/api/apm/search/:path*",
+        destination:
+          "https://incident-commander.canary.lab.flanksource.com/apm/search/:path*"
+      },
+      {
+        source: "/api/api/:path*",
```
moshloop, line 32: Why /api/api ?  whats the different between this and /api/canary ?

next.config.js
```diff
@@ -0,0 +1,44 @@
...
+        destination:
+          "https://incident-commander.canary.lab.flanksource.com/config/db/:path*"
+      },
+      {
+        source: "/api/canary/:path*",
+        destination:
+          "https://incident-commander.canary.lab.flanksource.com/canary/:path*"
+      },
+      {
+        source: "/api/incidents_db/:path*",
```
moshloop, line 22: Maybe /api/db/ ?

next.config.js
ciju: it's just mapping to existing paths. E.g. canary backend will receive /api/canary. The `api` prefix is for front-end and ui-backend to differentiate api calls. E.g. `/api/api` means canary request. etc.

next.config.js
ciju: There is also config_db, and canary etc. Can rename if it still makes sense. Thought it was better to be explicit about which db it is.

Dockerfile
ciju: Got it. Will remove from here.

moshloop:
@ciju can you rebase ?

moshloop:
@ciju apologies I should have merged this first, and now there is a merge conflict

PR #401, opened 2022-08-29
fix: increased debounce time of incident search input & handled stopping unnecessary calls

Prashanth-Nelli:


src/components/Dropdown/DropdownWithActions.tsx
```diff
@@ -1,112 +1,119 @@
...
+        let res = (await onQuery(query)) || [];
+        if (!res.length) {
+          setLastNoResultsQuery(query);
+        } else {
+          setLastNoResultsQuery("");
+        }
+        callback(res);
+      }
+      fetch();
+    }, 1000),
```
moshloop, line 57:
```suggestion
    }, 100),
```

moshloop:
@Prashanth-Nelli it doesn't look like the hypothesis dropdown updates when selecting a new / different incident

Prashanth-Nelli:
> @Prashanth-Nelli it doesn't look like the hypothesis dropdown updates when selecting a new / different incident

Done

PR #402, opened 2022-08-30
Canary - Filters, Labels with a single non-boolean value don't show

serberm:
Canary - Filters, Labels with a single non-boolean value don't show the value

### DIFF
--- a/src/components/Canary/TristateLabelStandalone/index.js
+++ b/src/components/Canary/TristateLabelStandalone/index.js
@@ -1,0 +1,47 @@
+import { useEffect, useState } from "react";
+import { TristateToggle } from "../../TristateToggle";
+import { updateParams, decodeUrlSearchParams } from "./../url";
+import { getConciseLabelState } from "./../labels";
+
+const TristateLabelStandalone = ({ label, className, labelClass, ...rest }) => {
+  const { labels: urlLabelState = {} } = decodeUrlSearchParams(
+    window.location.search
+  );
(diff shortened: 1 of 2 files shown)


moshloop:
So the issue is the technology tristate toggle should include the value (springboot)

serberm:
Could you provide more details?

src/components/Canary/TristateLabelStandalone/index.js
```diff
@@ -0,0 +1,47 @@
+import { useEffect, useState } from "react";
+import { TristateToggle } from "../../TristateToggle";
+import { updateParams, decodeUrlSearchParams } from "./../url";
+import { getConciseLabelState } from "./../labels";
+
+const TristateLabelStandalone = ({ label, className, labelClass, ...rest }) => {
+  const { labels: urlLabelState = {} } = decodeUrlSearchParams(
+    window.location.search
```
ciju, line 8: Could we use react-router [useSearchParams](https://reactrouter.com/en/main/hooks/use-search-params) for url params?

src/components/Canary/TristateLabelStandalone/index.js
```diff
@@ -0,0 +1,47 @@
...
+      onChange={(v) => handleToggleChange(v)}
+      className={className}
+      labelClass={labelClass}
+      label={label}
+      {...rest}
+    />
+  );
+};
+
+export default TristateLabelStandalone;
```
ciju, line 47: For new files, I would suggest using typescript.

PR #403, opened 2022-08-30
Canary Popup - Move selected check and modal state to the URL

serberm:


src/components/Canary/minimal.js
```diff
@@ -20,23 +21,39 @@ const MinimalCanaryFC = ({
   );
 
   useEffect(() => {
-    history.listen(({ location }) => {
+    const unlisten = history.listen(({ location }) => {
       setSearchParams(getParamsFromURL(location.search));
```
ciju, line 25: It's existing code, but since changes are related, consider replacing the logic with react-router useSearchParams?

PR #410, opened 2022-09-06
400 switch all dropdowns to react select + default to popout

Prashanth-Nelli:


src/components/AttachEvidenceDialog/index.tsx
```diff
@@ -349,7 +350,7 @@ export function AttachEvidenceDialog({
                     <span className="text-sm font-bold text-gray-700 mb-1 mr-4 w-16">
                       Type
                     </span>
-                    <Dropdown
+                    <ReactSelectDropdown
```
moshloop, line 353: Shouldn't we just change Dropdown to to use react-select ?

Prashanth-Nelli: I built a new component keeping in mind that in future if we want to go back previous implementation of Dropdown we can just replace it. If we modify the internal logic then it will complete re write again.

Prashanth-Nelli: I don't have a strong opinion around having it as a separate component. Let me know if you think there is no chance of going back to previous version of Dropdown, I'll just replace existing Dropdown with new implementation

PR #420, opened 2022-09-08
fix: fixed checks pop up data population issue

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli If I extend the time range, and then click on one of the junit-fail checks their is an unhandled exception

PR #432, opened 2022-09-12
feat: add gauge component

### LINKED ISSUE #133: Gauge Component
Create a component that can render a gauge e.g. for Memory / CPU isage
```json
{
label: "CPU",
value: 10,
max: 100
}
```

PulkitBanta:
### Short Description
Add a gauge component using `d3-shape` & `d3-scale`.
Label can be passed externally else the label will be made with `value` and `units`

## Example
```tsx
<Gauge
  value={67}
  minValue={0}
  maxValue={100}
  arcColor="green"
  arcBgColor="#d4d4d4"
  label={
    <h3 className="font-semibold text-xl text-center -mt-6">67%</h3>
  }
/>
```

## Screenshot
Screenshot 2022-09-12 at 4 19 20 PM

### Fixes #133


moshloop:
@PulkitBanta - Can we switch this to ant design charts? I think there are some issues with this components license.

PulkitBanta:
@moshloop Creating this with chart library will not be possible as in this case I am directly creating some svg elements while in chart libraries charts are created using some configurations.

As of License, these libraries can be used for commercial use.

src/components/Gauge/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    );
+  }
+);
+
+Gauge.defaultProps = {
+  units: "",
+  width: "12em",
+  arcColor: "green",
+  arcBgColor: "#d4d4d4"
+};
```
ciju, line 77: `defaultProps` is deprecated. And even otherwise, would be easier to set defaults in the params to the component.

PulkitBanta: Changed it to use defaults in the params only. Thanks

moshloop:
@PulkitBanta can you rebase and then squash your commits ?

PulkitBanta:
@moshloop I have squashed and rebased the changes

PR #433, opened 2022-09-12
409 new hypothesis created from attach evidence dialog are not show on incident page

Prashanth-Nelli:


moshloop:
It seems that the issue type and severity disappears when creating a new incident?

Prashanth-Nelli:
> It seems that the issue type and severity disappears when creating a new incident?

type & severity fields are being shown only during the new incident creation. if the user is selecting already created incident we are hiding these fields. 

There is any issue with the condition I have fixed it now.

PR #435, opened 2022-09-13
feat: add external id link in the responders list & tooltip

Prashanth-Nelli:


moshloop:
Screenshot 2022-09-13 at 17 22 11

- [ ] Can we change the remove to a trash icon on hover like the hypthosis?
- [ ] The external id in the popup should be a link with the id as the text, not the full URL
- [ ] For long ids (> 10 characters) use a prefix based ellipses with the full ID on hover

PR #449, opened 2022-09-16
Health-page-refresh-button

### LINKED ISSUE #437: Auto Refresh should be configurable
Screenshot 2022-09-13 at 18 02 59

On the health page, change the refresh button to look more like a button, with a separate dropdown, with the following values:

* None
* 15s
* 30s
* 60s
* 2m
* 3m
* 5m
* 10m
* 15m

- [ ] The selected value should be saved in local storage
- [ ] When the time range is greater than > 1h, disable the 15s, 30s options, if >3h disable 2m and 3m options
- [ ] If already set to a disabled option, auto-select the next interval ( only save it in local storage on user selection, though)
- [ ] Create a RefreshDropdown component so that it can be re-used

mainawycliffe:
Work in progress

Closes #437


mainawycliffe:
Closing due to commit mixe up

PR #450, opened 2022-09-16
feat: add RefreshDropdown Component

### LINKED ISSUE #437: Auto Refresh should be configurable

mainawycliffe:
Work in progress

Closes #437


moshloop:
@mainawycliffe I picked up a couple of issues

- [ ] The selected refresh period is not saved across different tabs / window reloads
- [ ] Updating the timePeriod disables the refresh period selections, but it doesn't change the selection to the first available, 
- [ ] The 3 dropdowns are not aliged / same height

mainawycliffe:
Will take care of this as soon as possible

moshloop:
@mainawycliffe it looks like only the first item was fixed:

Screenshot 2022-09-19 at 16 33 11

mainawycliffe:
@moshloop that's weird, let me take a closer look at that with a different monitor

mainawycliffe:
@moshloop I think I am done, can you take a look and let me know if there are other changes

mainawycliffe:
@moshloop I wired that to the old functionality of the refresh buttons, which I didn't touch, it updates the query params on click and is debounced. May be there is a bug there, let me investigate and see.

mainawycliffe:
@moshloop Upon further investigations, it doesn't seem like the function to refresh (as per my understanding of the codebase so far) is reacting to an update to the URL query params; this may be the reason why the refresh button isn't working. May be this would be a separate PR to fix this and I can take it.

moshloop:
@mainawycliffe Can you do some more testing, there are still a couple of issues:

* Clicking on the refresh button has a delay before the refresh starts
* The button doesn't have any hover styles
* It looks like the timers are overlapping / not getting cancelled - I can see multiple requests for different durations, some right after each other, some randomly.  When selecting none, it is still refreshing

mainawycliffe:
@moshloop the reason for the delay on clicking refresh button is because the fetch function is debounced, not sure why though

mainawycliffe:
@moshloop I have refactored the Component a little to make abort requests on subsequent, removed the delay on refresh button click, and added a hover style on the button.

moshloop:
@mainawycliffe - The button and aborting works - but it seems the background refresh isn't - Playing around with the intervals and refresh rates its pretty easy to get it to stop refreshing

mainawycliffe:
@moshloop let investigate this behavior

mainawycliffe:
@moshloop I think I have managed to track down the issue, the refresh would stop after first interval was done and I have fixed that.

src/types/healthChecks.ts
```diff
@@ -0,0 +1,52 @@
...
+  p95?: number;
+  rolling1h: number;
+}
+
+export interface HealthCheckUptime {
+  passed: number;
+  failed: number;
+}
+
+export interface HealthCheckLabels {
```
moshloop, line 44: This interface is incorrect, labels can be anything

mainawycliffe: You mean any key value map?

mainawycliffe: Done, let me resolve the merge conflcts

mainawycliffe:
@moshloop I think I have resolved the issue with the refresh interval not working now, when time range changes

mainawycliffe:
@moshloop can we merge this PR? It's been open for a long time and I am using it as a bases for my next task as it affects the pages in question, without repeating the work.

PR #452, opened 2022-09-16
feat: Component ordering

### LINKED ISSUE #430: Component ordering

aryaminus:
# CHANGELOG

- [x] Refactor the Topology Page and update responsiveness to handle sort button
- [x] Add Popover Sorting UI based on Name, Type, Last Updated and Headline Properties
- [x] Add click events to the sort filter button and popover buttons
- [x] Implement sorting the topology from the client side
- [x] Handle query params for updating the sort object
- [x] Switch to BsSortDown and BsSortUp icons and make the field name visible so that you don't need to click on the dropdown to see what the sort field is
- [x] To change the sort order, either click on the icon or click on the field in the drop-down again
- [x] The cpu/memory ordering doesn't look like it is working - we probably need custom sort functions for property type, millicores, bytes, numbers, etc.. (string, number and date are handled, so it should work)
- [x] Add in health as the default sort (desc) with info = 0, healthy = 1, warning = 2, unhealthy = 3 - When this sort is selected it should remove the params for the url
- [x] Follow commit conventions of the repo
- [x] Check for side effects

Closes #430


moshloop:
@aryaminus thanks this is a good start, just a couple of minors:

- [ ] Switch to BsSortDown and BsSortUp icons and make the field name visible so that you don't need to click on the dropdown to see what the sort field is
- [ ] Tochange the sort order, either click on the icon or click on the field in the drop-down again
- [ ] The cpu/memory ordering doesn't look like it is working - we probably need custom sort functions for property type, millicores, bytes, numbers, etc..
- [ ] Add in health as the default sort (desc) with info = 0, healthy = 1, warning = 2, unhealthy = 3 - When this sort is selected it should remove the params for the url

aryaminus:
@moshloop Do let me know how the selected field name UI looks. Looks a bit off

moshloop:
@aryaminus Please check 

- [ ] Build errors
- [ ] squash your commits / recommit - do not add icons into the commit message
- [ ] When click on the sort icon it should flip sort direction

aryaminus:
@moshloop Cherrypicking the commits in a new PR

PR #454, opened 2022-09-19
feat: Component ordering

### LINKED ISSUE #430: Component ordering

aryaminus:
# CHANGELOG

- [x] Refactor the Topology Page and update responsiveness to handle sort button
- [x] Add Popover Sorting UI based on Name, Type, Last Updated and Headline Properties
- [x] Add click events to the sort filter button and popover buttons
- [x] Implement sorting the topology from the client side
- [x] Handle query params for updating the sort object
- [x] Switch to BsSortDown and BsSortUp icons and make the field name visible so that you don't need to click on the dropdown to see what the sort field is
- [x] To change the sort order, either click on the icon or click on the field in the drop-down again
- [x] The cpu/memory ordering doesn't look like it is working - we probably need custom sort functions for property type, millicores, bytes, numbers, etc.. (string, number and date are handled, so it should work)
- [x] Add in health as the default sort (desc) with info = 0, healthy = 1, warning = 2, unhealthy = 3 - When this sort is selected it should remove the params for the url
- [x] Follow commit conventions of the repo
- [x] Cherrypick from https://github.com/flanksource/flanksource-ui/pull/452
- [x] Build errors
- [x] Introduce Popover to handle Preferences and Sorting

Closes #430


src/pages/TopologyPage.tsx
```diff
@@ -257,137 +358,220 @@ export function TopologyPage() {
...
+                  onClick={() => onSelectSortOption(sortBy, "desc")}
+                />
+              )}
+              {sortByType === "desc" && (
+                <BsSortUp
+                  className="w-6 h-6 text-gray-700 hover:text-gray-900"
+                  onClick={() => onSelectSortOption(sortBy, "asc")}
+                />
+              )}
+              <span
```
moshloop, line 474:
```suggestion
                <BsSortUp
                  className="w-6 h-6 text-gray-700 hover:text-gray-900"
                  onClick={() => onSelectSortOption(sortBy, "desc")}
                />
              )}
              {sortByType === "desc" && (
                <BsSortDown
                  className="w-6 h-6 text-gray-700 hover:text-gray-900"
                  onClick={() => onSelectSortOption(sortBy, "asc")}
                />
              )}
              <span
```

src/components/IncidentDetails/index.tsx
```diff
@@ -1,30 +1,32 @@
 import { useEffect, useMemo, useState } from "react";
-import { useForm } from "react-hook-form";
+
```
moshloop, line 2: What are the changes in this file for ?

aryaminus: There were build errors in `main` branch. Just so happened that this PR and another PR addressed it.
The changes in this file is just ordering done by `Tailwind CSS IntelliSense` vscode extension. Would you prefer to revert?

aryaminus:
> @aryaminus Can we extract the sort dropdown and preferences dropdown into standalaone components ?

@moshloop Can do. We could just create a `Popover` component and handle dynamic component. I'll refactor with that in mind

moshloop:
@aryaminus So a popover component is a good idea, but I also want to split this file up as it is getting very large

src/components/TopologyPopover/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+            onClick={() => onSelectSortOption(sortBy, "asc")}
+          />
+        )}
+        <span
+          className="hidden ml-2 text-base text-gray-700 capitalize bold md:flex hover:text-gray-900"
+          onClick={() => setCurrentIcon((val) => (val === "" ? "sort" : ""))}
+        >
+          {sortTypes.find((s) => s.value === sortBy)?.label}
+        </span>
+      </div>
```
moshloop, line 136: shouldn't this be in topologySort.tsx ? 

I am expecting just 
```jsx
<TopologySort
     topology={topology}
     sortBy={sortBy}
     sortDir={sortDir}
     onSelect={onSelectSortOption} />
```

src/components/TopologyPopover/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  const [sortTypes, setSortTypes] = useState<typeof defaultSortTypes>([]);
+
+  useEffect(() => {
+    setSortValues();
+  }, [searchParams, topologyLabels]);
+
+  useEffect(() => {
+    const listener = (event: MouseEvent) => {
+      if (currentIconRef.current?.contains(event.target)) {
```
moshloop, line 53: not quire sure I understand what the purpose of currentIcon is for ?

src/components/TopologyPopover/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+};
+
+export type Topology = {
+  id: string;
+  name: string;
+  type?: string;
+  updated_at?: string;
+  properties?: TopologyProperty[];
+};
+
```
moshloop, line 19: I don't think these should be here

src/components/TopologyPopover/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  if (sortByType === "desc") {
+    return updatedTopology.reverse();
+  }
+
+  return updatedTopology;
+}
+
+export function getCardWidth() {
```
moshloop, line 96: shouldn't this be in the preferences ?

src/components/TopologyPopover/index.tsx
aryaminus: The plan is to have both `TopologyPreference` and `TopologySort` to be stateless. So, props are drilled down to handle any kind of events ie. Menu icon and Events inside Popover

moshloop: Why is there any sharing required? The preference and sort popovers should maintain their own state in their own popover.

aryaminus: Primarily to have a single handler when clicking outside the active popover. I'll create separate refs and send a commit

PR #455, opened 2022-09-20
feat: add support for fixed values & defaults in responder form

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli How do use values from the incident in the templating like title, type, commander, etc..

Prashanth-Nelli:
> @Prashanth-Nelli How do use values from the incident in the templating like title, type, commander, etc..

@moshloop currently templates are resolved using the responder form values. If we need to support other values we need to add these values as key value pair in the context object. 

eg:-

const context = {
   incident: 'Test Incident',
   title: 'Test title',
   commander: 'Adam',
   ...{responder form values}
};

defaultValues: {
   summary: 'This is related to ${incident} with ${title} created by ${commander}'
}

- and currently templates are only support in default values. For fixed values as we are disabling the edits it doesn't make sense to have templating.

moshloop:
@Prashanth-Nelli  - it doesn't seem to be working for me, the project is pre-populated but the others are not ? Also the severity is not editable, but I don't see it as a values ?

PR #456, opened 2022-09-20
feat: display date as per user locale 

### LINKED ISSUE #353: Ensure consistent timestamp handling
* Should use the same format. 
* Assume UTC and convert to local timezone

PulkitBanta:
### Description
Timestamp values are formatted locally as per the user while being displayed.

### Fixes #353


src/components/Canary/CanaryStatusChart/index.js
```diff
@@ -15,6 +14,7 @@ import { getCanaryGraph } from "../../../api/services/topology";
 import { Loading } from "../../Loading";
 import { useSearchParams } from "react-router-dom";
 import { getParamsFromURL } from "../utils";
+import { formatDate } from "../../../utils/formatDate";
```
moshloop, line 17: Can we extract all of the date-related functionality into utils/date.js ? any use of dayjs etc should be isolated to that file.

src/components/Hypothesis/EvidenceSection/index.tsx
```diff
@@ -65,7 +65,7 @@ const EvidenceAccordion: React.FC<{
             {title || <span className="text-gray-400">(no title)</span>}
             {date && (
               <div className="text-gray-400 text-sm">
-                {dayjs(date).format("HH:mm A, MMM DD, YYYY")}
+                {formatDate(date, "HH:mm A, MMM DD, YYYY")}
```
moshloop, line 68: The data format should be in date.js -  different components shouldn't be using different formats.  it should be one:

* relative
* long (textual)
* timestamp 
* time

moshloop:
I think incident details table also needs to be catered for

moshloop:
I still don't see a change on incidents.

src/components/Canary/CanaryStatusChart/index.js
```diff
@@ -55,11 +55,11 @@ export function CanaryStatusChart({ check, ...rest }) {
   }
 
   // @TODO: date should be formatted properly depending on selection, not just by DD/MM
-  let formatDate = (date) => dayjs(date).format("HH:mm");
+  let customFormatDate = (date) => formatDate(date, "HH:mm");
```
moshloop, line 58: formatDate should not accept a format - formats can only be defined in date.ts, and they must be consistently used. 
In this case relativeDate is probably better

src/components/Canary/CanaryStatusChart/index.js
PulkitBanta: Sure, I will make that change

src/components/Canary/CanaryStatusChart/index.js
PulkitBanta: I am defining a enum for the date format types which will be consistent and is going to have 3-4 types which will be used throughout the application

src/components/Canary/CanaryStatusChart/index.js
```diff
@@ -15,6 +14,7 @@ import { getCanaryGraph } from "../../../api/services/topology";
 import { Loading } from "../../Loading";
 import { useSearchParams } from "react-router-dom";
 import { getParamsFromURL } from "../utils";
+import { DATE_FORMATS, formatDate } from "../../../utils/date";
 
 // @TODO: duration should be formatted properly, not just by ms
 const formatDuration = (duration) => `${duration}ms`;
```
moshloop, line 20: this formatDuration method itself should move, not just the formats

Prashanth-Nelli:
closing this PR as the requirement is covered in https://github.com/flanksource/flanksource-ui/pull/559

PR #461, opened 2022-09-22
chart config to set NEXT_PUBLIC_WITHOUT_SESSION

ciju:


chart/values.yaml
```diff
@@ -15,6 +15,9 @@ nameOverride: "flanksource-ui"
 fullnameOverride: "flanksource-ui"
 oryKratosURI: ""
 
+# Option to disable auth on frontend.
+withoutOrySession: true
```
moshloop, line 19: Can we rename this to just authentication?

ciju: change to `withoutAuth`.

PR #462, opened 2022-09-22
feat: attach health check to incident

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli can we

- [ ] Move the attach button into (a new ) footer
- [ ] On the incident page, don't show the graph, maybe just the dots, latency, duration - and then on click popup full window.

moshloop:
@Prashanth-Nelli the view is too large, I am expecting 1 row for the status and 1 row for details - 

Icon Name                                 * * * * *  Uptime: 99%  Latency: 55ms
[ Some long description ]

Prashanth-Nelli:
> @Prashanth-Nelli the view is too large, I am expecting 1 row for the status and 1 row for details -
> 
> Icon Name * * * * * Uptime: 99% Latency: 55ms [ Some long description ]

Screenshot 2022-09-27 at 10 57 56 PM

is this view okay ?, I'll add recent logs details in the bottom as a one liner.

Prashanth-Nelli:
we support two views

1. large view
Screenshot 2022-09-27 at 11 00 18 PM
2. small view
Screenshot 2022-09-27 at 10 57 56 PM

moshloop:
@Prashanth-Nelli  except the table doesn't really make sense as it will be a single row only - so the table headers should move inline into the row.

Prashanth-Nelli:
> @Prashanth-Nelli except the table doesn't really make sense as it will be a single row only - so the table headers should move inline into the row.

Screenshot 2022-09-28 at 4 33 23 PM

made changes as suggested.

PR #463, opened 2022-09-22
Config analysis changes

moshloop:


.env
```diff
@@ -2,4 +2,5 @@ SKIP_PREFLIGHT_CHECK=true
 PORT=3004
 # ORY_KRATOS_URL=http://kratos.canary.lab.flanksource.com/
 # NEXT_PUBLIC_APP_DEPLOYMENT=CANARY_CHECKER
-# NEXT_PUBLIC_WITHOUT_SESSION=true
+NEXT_PUBLIC_WITHOUT_SESSION=true
+BASE_URL=https://incident-commander.canary.lab.flanksource.com/
```
ciju, line 6: For local develpment, these environment variables could be set at .env.local. Otherwise these values would be default for production build. .env.local is gitignored.

PR #467, opened 2022-09-25
feat: Canary Status Chart - Configurable time range

### LINKED ISSUE #209: Canary Status Chart - Configurable time range
- Provide configurable range for canary chart (The chart is in the dialog shown when click on a row in the Health table)
 - Save selected range back to the search URL

Screenshot 2022-04-17 at 20 46 19

aryaminus:
## CHANGELOG

- [x] ~~Introduce basic time buttons above the chart~~
- [x] ~~Reuse the DateTimePicker for range based calculations~~
- [x] Reuse the Dropdown to match the query param and update chart
- [x] Handle updating the range object and re-render
- [x] Improve UI
- [x] Update the X-Axis based on Range
- [x] Preserve query param for check dialog

Closes #209


aryaminus:
@moshloop I reused the dropdown instead of tabs to handle responsiveness and utilize the same query params. Do you want me to revert back to tabs as shown in screenshot?

moshloop:
@aryaminus a couple of issues I can see:

* The values in the chart don't seem to correspond to the values from the API
Screenshot 2022-09-28 at 15 14 10
* When the dialog is opened, I expect the time range to be the same as the time range specified on the main page
* When changing the time range in the chart, it should not change the time range of the main page 
* The state should be stored in the URL, so that I can copy and paste the url, and then it will automatically open the dialog,  with the selected check, and time range

aryaminus:
> @aryaminus a couple of issues I can see:
> 
> * The values in the chart don't seem to correspond to the values from the API
> 
> Screenshot 2022-09-28 at 15 14 10
> 
> * When the dialog is opened, I expect the time range to be the same as the time range specified on the main page * When changing the time range in the chart, it should not change the time range of the main page * The state should be stored in the URL, so that I can copy and paste the url, and then it will automatically open the dialog, with the selected check, and time range

@moshloop I modified the API to reduce the array and concatenate feature sets. For example: For hrs to 1 day, it concatenates to 00:00-24:00, For 2months to 1 Year, it concatenates to Jan-Dec

moshloop:
> I modified the API to reduce the array and concatenate feature sets.

Why ?  Those data points are needed - Rendering 300 data points on a graph should not be a performance issue.  

Eventually, we may use sampling on the server-side to reduce the data points, but from a client side it should render what is returned

aryaminus:
> > I modified the API to reduce the array and concatenate feature sets.
> 
> Why ? Those data points are needed - Rendering 300 data points on a graph should not be a performance issue.
> 
> Eventually, we may use sampling on the server-side to reduce the data points, but from a client side it should render what is returned

@moshloop Not a performance issue. It was just to collate a make singular data points based on date. Reverted

PR #471, opened 2022-09-28
Config updates

moshloop:


src/components/AttachEvidenceDialog/index.tsx
```diff
@@ -269,7 +278,13 @@ export function AttachEvidenceDialog({
     }
 
     if (isNewHypothesis) {
-      const nodeDetails = { type: "root" };
+      /* type should be root, if it's the first for an incident? */
+      const nodeDetails = isNewIncident
+        ? { type: "root" }
+        : {
+            type: "factor",
```
Prashanth-Nelli, line 285: @moshloop are we going to have only one root hypothesis or multiple ones ?

PR #472, opened 2022-09-29
ui and build fixes for standalone deployment

ciju:


src/components/Canary/index.tsx
```diff
@@ -171,7 +171,7 @@ export function Canary({
   }
 
   return (
-    <div className="flex flex-row place-content-center">
+    <div className="flex flex-row place-content-center h-screen overflow-y-auto">
```
ciju, line 174: @Prashanth-Nelli added this for scroll not happening on canary checker standalone app. I think you had done some related changes. Please update this, if there is a better fix.

PR #482, opened 2022-10-04
feat: health page filters relayout

Prashanth-Nelli:


moshloop:
It looks like there is a z-index issue - I think we should switch to the new timerange component anyway

Screenshot 2022-10-06 at 09 27 05

Prashanth-Nelli:
> It looks like there is a z-index issue - I think we should switch to the new timerange component anyway
> 
> Screenshot 2022-10-06 at 09 27 05

Screenshot 2022-10-10 at 12 06 39 PM

fixed

Issue #422, opened 2022-09-08
Resource Editor - Default to YAML


moshloop:
Convert to and from JSON when saving, but allow editing in YAML 

This applies to Settings -> Teams/Incident Rules/Etc

PR #492, opened 2022-10-12
feat: add yaml as default for resource editor

sagarrchauhan:
### Description

Added feature to show YAML as default for resource editor. Data is still saved and retrieved as JSON.

PR references to #422


src/components/SchemaResourcePage/SchemaResource.tsx
```diff
@@ -25,7 +27,11 @@ export function SchemaResource({
     if (!id) return;
     setLoadingState("loading");
     getResource(resourceInfo, id)?.then((res) => {
-      setResource(res.data[0]);
+      const specData = res.data[0]?.spec;
+      setResource({
+        ...res.data[0],
+        spec: specData && !isEmpty(specData) ? dump(specData) : ""
```
moshloop, line 33: can we set sortKeys = true?

sagarrchauhan: I did not understand the context. Do you want me to add a property called sortKeys to the resource object?

moshloop: No - js-yaml supports an option called `sortKeys` when dumping

sagarrchauhan: Added the option

PR #493, opened 2022-10-12
fix: components filtering issue fix

Prashanth-Nelli:


src/pages/TopologyPage.tsx
```diff
@@ -354,6 +363,36 @@ export function TopologyPage() {
             ).map((item) => (
               <TopologyCard key={item.id} topology={item} size={size} />
             ))}
+            {!topology?.length && (
+              <div className="w-full flex justify-center">
+                <div className="w-96 mt-16">
+                  <div className="rounded-md bg-blue-50 p-4">
+                    <div className="flex">
+                      <div className="flex-shrink-0">
+                        <svg
```
moshloop, line 372: Why is this inlined like this ?

Prashanth-Nelli: @moshloop  I have used tailwindcss alert templates directly, I'll replace the svg with a similar icon component.

PR #495, opened 2022-10-12
feat: add side panel for topology page

### LINKED ISSUE #481: Topology Detail v2.0
Master-Detail Cards

mainawycliffe:
Closes #481


mainawycliffe:
@moshloop this is almost ready; just one thing remaining in terms of Topology Details: We have the icon section and list below. Do we want to exclude the ones with a headline set true from the details list and only show them on the icon section? Secondly, where do the icons for the icon section come from?

moshloop:
> Do we want to exclude the ones with a headline set true from the details list and only show them on the icon section?

yes

> where do the icons for the icon section come from?

these would come from the property.icon field

mainawycliffe:
Thanks. I will take care of those.

moshloop:
@mainawycliffe  The metrics are still missing: 

Screenshot 2022-10-14 at 11 27 43

PR #496, opened 2022-10-12
feat: display dates as per user locale

sagarrchauhan:
Added changes from PR #456 

Fixes issues in #353


src/utils/date.ts
```diff
@@ -10,4 +11,127 @@ export function isDate(dateString?: ValueType) {
...
+};
+
+/**
+ * Parse the given date relative to the user locale
+ *
+ * @param date
+ * @param fromNow give relative date from now
+ * @returns relative date as string
+ */
+export const relativeDateTime = (date: string | Date, fromNow = false) => {
```
moshloop, line 56: Why would a relative date be anything from now ?

src/utils/date.ts
```diff
@@ -10,4 +11,127 @@ export function isDate(dateString?: ValueType) {
...
+ */
+export const dateToJSDate = (value: DateConfig) => dayjs(value).toDate();
+
+/**
+ *
+ * @param value date to format
+ * @param format custom string format for the date
+ * @returns formatted string of the date
+ */
+export const customDateFormattor = (value: DateConfig, format: string) =>
```
moshloop, line 99: Why is this needed ?

src/utils/date.ts
```diff
@@ -10,4 +11,127 @@ export function isDate(dateString?: ValueType) {
...
+) => dayjs(date1).diff(dayjs(date2), diffUnit as ManipulateType, isDiffInFloat);
+
+/**
+ *
+ * @param date1 First date
+ * @param date2 Second Date
+ * @param withoutSuffix If suffix needed at the end of time passed.
+ * @returns a string mentioning the time passed from date2 relative to date1
+ */
+export const timePassedInWords = (
```
moshloop, line 133: how is this different from relativeDateTime?

src/utils/date.ts
```diff
@@ -10,4 +11,127 @@ export function isDate(dateString?: ValueType) {
...
+ */
+export const dateToISOString = (value: DateConfig) =>
+  dayjs(value).toISOString();
+
+/**
+ *
+ * @param value date to convert
+ * @returns Native date object
+ */
+export const dateToJSDate = (value: DateConfig) => dayjs(value).toDate();
```
moshloop, line 91: What is this need for? shouldn't it be parseDate ?

src/utils/date.ts
sagarrchauhan: @Prashanth-Nelli has used a lot of custom formats for displaying dates in getUpdatedFormat() function of src/components/Canary/CanaryStatusChart/index.tsx. These formats are different from the ones mentioned in https://github.com/flanksource/flanksource-ui/pull/456#discussion_r976382121. 

Hence I added a separate option for displaying these custom formats. Let me know if we there is a different requirement here.

src/utils/date.ts
sagarrchauhan: This function is being used to get native date object from either string, dayjs object, milliseconds etc. 

We can use normal parsing with Date object constructor, but in case it is a dayjs object then it is better to handle it with dayjs library. I have used this as a global function for any kind of input. If we need to change this, then let me know.

src/utils/date.ts
sagarrchauhan: The function is not exactly giving only the relative date. The way it was originally written , it either gives the duration passed from now for the "date" passed in the parameter (if "fromNow" is set to true). If "fromNow" is not specified, it gives duration only if the "date" parameter passed is the same day as today. Else it gives the date-time format as specified but with respect to the user local. It is being used in a lot of places, so I assumed the working of the function is right.

Let me know if we need to make changes to this function.

src/utils/date.ts
sagarrchauhan: The functionality here is totally different.

1. It takes 2 dates as params, rather than 1 for measuring the time different between the two. This is because the second date is not necessarily "now".
2. relativeDateTime does a custom operation as I have mentioned in https://github.com/flanksource/flanksource-ui/pull/496#discussion_r1006519326. This functionality is very different from what is being done here.

Let me know if there is a need to have some separate functionality here.

src/utils/date.ts
moshloop: @sagarrchauhan we want to standardize on date formats, not have different components using different formats

src/utils/date.ts
moshloop: We don't want to have inconsistent dates, relalativeDate should always return the duration from now

Issue #473, opened 2022-09-29
Config List grouping


moshloop:
- Type (default)
- Analysis
- Changed (time buckets of today, this week etc)
- Account

PR #506, opened 2022-10-14
feat: add group by option for data table

sagarrchauhan:
To add the feature for grouping in Config List. Also moved header items in side the body
Addresses: #473

### DIFF
--- a/src/components/ConfigChangeHistory/index.jsx
+++ b/src/components/ConfigChangeHistory/index.jsx
@@ -1,4 +1,4 @@
-import { Link } from "react-router-dom";
+import { Link, useSearchParams } from "react-router-dom";
 
 import { DateCell } from "../ConfigViewer/columns";
 import { DataTable } from "../index";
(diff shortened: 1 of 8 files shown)


moshloop:
@sagarrchauhan  

- [ ] We seem to have lost the icons
- [ ] The +/- is on its own line
- [ ] The click target should be the whole cell
- [ ] Use the same icon as for the tags column

sagarrchauhan:
Have addressed the issues:

- [ ] We seem to have lost the icons - Restored them.
- [ ] The +/- is on its own line - Added a different chevron item which is in line with the item.
- [ ] The click target should be the whole cell - Made this clickable
- [ ] Use the same icon as for the tags column - Used the icons as per the Icons function.

moshloop:
@sagarrchauhan it looks like the column sizing has been changed to equal width - it should shrink to width, with a maximum width, the name column should also ideally be underneath the group, not next to

i.e. 

Instead of :
```
EC2Instance
                         Name123

EC2Instance (5)
      Name 1
      Name 2
```

moshloop:
@sagarrchauhan  - thanks looking better, still outstanding:

* Alignment of name column 
* Count of items in brackets next to type
* Total of changes ( and the cost columns  now)
*

sagarrchauhan:
@moshloop, I have addressed the issues:

- Alignment of name column - Have aligned it below the group name
- Count of items in brackets next to type - Added Badge for showing the count
- Total of changes ( and the cost columns now) - Made changes to Changed and cost columns by showing aggregate for them

moshloop:
@sagarrchauhan its looking better:

- Don't show $0.00 or 0 values
- Groups can't have created/updated dates
- The column sizing should not be equal, the columns should shrink to width, and then tags / analyis / changes max-width of 400
- The label for the grouping column is wrong
- There is no way to sort on the grouping column (it should sort by it A-Z by default)

sagarrchauhan:
I have addressed the following issues:

- Don't show $0.00 or 0 values - **Removed the 0 values**
- Groups can't have created/updated dates - **Showing "-" for created/updated dates in group**
- The column sizing should not be equal, the columns should shrink to width, and then tags / analyis / changes max-width of 400 -  **Initially on page load, the column size looks equal but it is not as there are very less values to show in groups. On expanding the group the column sizes change as per available content inside the groups. But then also, I have set max-width of 400 for required columns.** 
- The label for the grouping column is wrong - **Have fixed this**
- There is no way to sort on the grouping column (it should sort by it A-Z by default) - **By default the grouping is ascending as per grouping column.**

Prashanth-Nelli:
changes are covered in #554, so closing this PR

PR #508, opened 2022-10-17
fix: fixed topology creation issue from ui

Prashanth-Nelli:


src/components/SchemaResourcePage/SchemaResourceEdit.tsx
```diff
@@ -66,10 +72,24 @@ export function SchemaResourceEdit({
   }
   watch("spec");
 
+  const hasNamespaceField = useMemo(() => {
```
moshloop, line 75: Shouldn't this be part of the resource definition ?

Prashanth-Nelli: Yeah, that is the right place for this logic. I'll refactor this accordingly.

PR #511, opened 2022-10-18
feat: set flatten to true, when topology type is selected

### LINKED ISSUE #502: Topology query - When selecting a topology type, set flatten=true

mainawycliffe:
Closes #502


mainawycliffe:
Makes sense, let me fix that.

mainawycliffe:
@moshloop this should be good to go now

PR #514, opened 2022-10-18
fix: opened checks modal synchronously without waiting for api response

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli - I can still see some values (counts) from a previous check when selecting a large time range, and selecting multiple checks - I think we need to null out the selected value on modal close

Can we also merge in https://github.com/flanksource/flanksource-ui/pull/467 ?

moshloop:
Just an issue with the dropdown - Can this not fixed globally ?
Screenshot 2022-10-19 at 10 48 29

Prashanth-Nelli:
> Just an issue with the dropdown - Can this not fixed globally ? Screenshot 2022-10-19 at 10 48 29

This issue can't be fixed globally, as the content is getting clipped due to container's horizontal overflow property.

Prashanth-Nelli:
> > Just an issue with the dropdown - Can this not fixed globally ? Screenshot 2022-10-19 at 10 48 29
> 
> This issue can't be fixed globally, as the content is getting clipped due to container's horizontal overflow property.

I'll explore left aligned and right alined dropdown option. if the react select supports this feature then we can solve this basic prop changes.

PR #518, opened 2022-10-19
feat: add costs to config list and details

### LINKED ISSUE #515: Add costs
- On cards
- Conflig list

mainawycliffe:
Closes #515


moshloop:
@mainawycliffe 
- [ ] Apply the rounding rules as per: https://github.com/flanksource/flanksource-ui/blob/beb0513675b1ca63147697579e40336f5e45cee1/src/components/TopologyCard/Property.tsx#L28
- [ ] The column titles should be:  Cost (per min),  Cost (24h), Cost (7d),  Cost (30d) 
- [ ] In config detail, use $ icon rather than note
- [ ] In config details use stat styles i.e. large value, and small label underneath

moshloop:
Can we also

- [ ] Pin the table header to the top
- [ ] Fix the desc/asc icon - I think it is inverted
- [ ] In config detail, if there are no costs - dont show them

mainawycliffe:
I have taken care of the above issue

src/components/ConfigCosts/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (value > 1000) {
+      return (parsedValue / 1000).toFixed(1) + "k";
+    }
+    if (parsedValue > 100) {
+      return parsedValue.toFixed(0);
+    }
+    if (parsedValue > 10) {
+      return parsedValue.toFixed(1);
+    }
+    if (parsedValue > 0.01) {
```
moshloop, line 18: I think lets always round up to 2 decimal places

mainawycliffe: @moshloop done

moshloop:
@mainawycliffe - the round up to 2 decimal places is only for values < $10, for for values >$10 it should be no decimal places

For the cost could we use something like:

```html
<dl class="grid grid-cols-1 gap-5 sm:grid-cols-3">
    <div class="overflow-hidden  px-4 py-2  ">
      <dt class="truncate text-sm font-medium text-gray-500">Per Min</dt>
      <dd class="mt-1 text-xl font-semibold tracking-tight text-gray-900">$1</dd>
    </div>

    <div class="overflow-hidden px-4 py-5  ">
      <dt class="truncate text-sm font-medium text-gray-500">1d</dt>
      <dd class="mt-1 text-xl font-semibold tracking-tight text-gray-900">$100</dd>
    </div>

    <div class="overflow-hidden  px-4 py-5  ">
      <dt class="truncate text-sm font-medium text-gray-500">7d</dt>
      <dd class="mt-1 text-xl font-semibold tracking-tight text-gray-900">$500</dd>
    </div>
  </dl>
```

mainawycliffe:
@moshloop  Something like this?

image
image

PR #526, opened 2022-10-24
feat: add filter by component for incidents list

### LINKED ISSUE #425: Incidents - Filter by component

mainawycliffe:
Closes #425


mainawycliffe:

moshloop:
Yes - but switch to `component_names` rather than components - components table directly itself should never be called

mainawycliffe:
Cool...let me make the change

mainawycliffe:
@moshloop, it seems like the react-select is already set to popout for all controls

moshloop:
@mainawycliffe  when show the value, I am not expecting a border it should be without a border and shrink to content, The popover should have a min width, and expand up to 400px with the content - Switching the component to be the first filter may also help

mainawycliffe:
@moshloop I get you now.

moshloop:
I am not seeing any changes ?

mainawycliffe:
I haven't pushed the changes yet, still trying to localize the changes so they don't affect components in other places where were use the component. 
image

mainawycliffe:
Is this what you had in mind?
image

moshloop:
@mainawycliffe , yeh - Can we maybe move them right on top of the table like topology / health, and then order them in the same way the table columns are ordered ?

mainawycliffe:
@moshloop Yeah, let me do that.

mainawycliffe:
Done, at the moment, we don't have a column for components or owner, what do we do with those.

moshloop:
@mainawycliffe  both the filter bar and table headers should be fixed, only the rows should scroll

moshloop:
@mainawycliffe 

- when the page first loads, the filter bar and table is in the center of the page vs the top
- can we style the dropdowns like the health page group by / tab by i.e The label inside the field (excecpt the dropdown should still shrink to width

mainawycliffe:
@moshloop is this what you mean? 
image

moshloop:
@mainawycliffe remove the extra component label, make the background of the inputs white, and change the label text to gray-600

mainawycliffe:
@moshloop something like this? 
image

PR #528, opened 2022-10-25
feat: show toast animation when navigating topology

### LINKED ISSUE #505: Topology - Add top of page bouncing progress bar - full width
e.g. something like http://nanobar.jacoborus.codes/

mainawycliffe:
Closes #505

### DIFF
--- a/src/pages/TopologyPage.tsx
+++ b/src/pages/TopologyPage.tsx
@@ -24,6 +24,7 @@ import { getSortedTopology } from "../components/TopologyPopover/topologySort";
 import { getCardWidth } from "../components/TopologyPopover/topologyPreference";
 import TopologySidebar from "../components/TopologySidebar";
 import { BsFillInfoCircleFill } from "react-icons/bs";
+import toast from "react-hot-toast";
 
 const allOption = {
   All: {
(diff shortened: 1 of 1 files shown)


moshloop:
Why do we need this ? How is this different to the refresh button or the skeleton ?

mainawycliffe:
While topology is loading and there was another topology being displayed, we don't show the loader, we show the existing topology until the other one is loaded, and the user is left wondering whether clicking worked. So, we can either do what I have done here or just show a skeleton as soon as the user navigates to another topology

moshloop:
I don't see what this is doing? Can you record a video. 

I did not that the sidepanel loads, and then the topology cards are updated - We should probably only show the side panel when those details are retrieved, and empty the existing components before sending the load request (only for chaging the id, for all other filter changes the loading indicator at the top should be enough

mainawycliffe:
The animation is at the bottom center of the screen. As shown below:

https://user-images.githubusercontent.com/12270550/198031840-b7744961-3b3b-4e87-8d4e-3c3be1269c73.mp4

mainawycliffe:
> We should probably only show the side panel when those details are retrieved, and empty the existing components before sending the load request (only for chaging the id, for all other filter changes the loading indicator at the top should be enough

Yeah, this should work.

PR #529, opened 2022-10-25
feat: incident definition of done

Prashanth-Nelli:


moshloop:
- [ ] Can we remove the borders around the items and style it to look more like the responders list ?
- [ ] The checkbox indicates whether `done` is true, and `definition_of_done` indicates whether it should be in the list altogher, 
- [ ] when selecting Add definition of done, you can then select existing evidence items that are not marked definition_of_done

Prashanth-Nelli:
> * [ ]  Can we remove the borders around the items and style it to look more like the responders list ?
> * [ ]  The checkbox indicates whether `done` is true, and `definition_of_done` indicates whether it should be in the list altogher,
> * [ ]  when selecting Add definition of done, you can then select existing evidence items that are not marked definition_of_done

Currently I am using done flag to show hour glass icon & checkmark icon. which property in the evidence should I consider to show these icons.

src/components/Hypothesis/ResponseLine/index.tsx
```diff
@@ -74,17 +79,69 @@ export function ResponseLine({
...
-        />
+
+      {(markAsDefinitionOfDone || onDelete) && (
+        <Menu>
+          <Menu.VerticalIconButton />
+          <Menu.Items widthClass="w-72">
+            {onDelete && (
+              <Menu.Item onClick={() => Promise.resolve(onDelete())}>
+                <div className="cursor-pointer flex w-full">
+                  <IconButton
```
moshloop, line 90: Can we extract this into a component?

src/components/IncidentDetails/DefinitionOfDone/EvidenceView/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+        </div>
+        <div className={clsx("overflow-hidden")}>
+          <div className="flex flex-row">
+            <span
+              title={check?.name}
+              className={clsx(
+                "text-gray-800 font-semibold whitespace-nowrap overflow-ellipsis overflow-hidden pr-4"
+              )}
+            >
+              {check?.name}
```
moshloop, line 235: Why is this duplicayed ?

src/components/IncidentDetails/DefinitionOfDone/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+                          </div>
+                        </Menu.Item>
+                      </Menu.Items>
+                    </Menu>
+                  </div>
+                </div>
+              );
+            })}
+          <div className="flex items-center justify-between p-2 pl-1">
+            <button
```
moshloop, line 199: Extract this into a component

src/components/IncidentDetails/DefinitionOfDone/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+          className={clsx(
+            "flex rounded-t-lg justify-between bg-gray-100 px-8 pb-4 items-end",
+            "absolute w-full bottom-0 left-0"
+          )}
+        >
+          <div className="flex flex-1">
+            <button
+              type="submit"
+              className={clsx(
+                "inline-flex items-center px-4 py-2 border border-gray-300 shadow-sm font-medium rounded-md text-gray-700 bg-white hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-blue-500 sm:text-sm",
```
moshloop, line 317: use the standard button classes

src/components/IncidentDetails/DefinitionOfDone/EvidenceView/index.tsx
Prashanth-Nelli: I have created a simpler version here as the regular evidence views are heavy to render because the view contains code for other components and their supporting logic.

src/components/IncidentDetails/DefinitionOfDone/index.tsx
Prashanth-Nelli: I think we have used this way in multiple places. should I refactor there as well with btn-primary class

PR #541, opened 2022-10-28
feat: api query caching

Prashanth-Nelli:


src/api/query-hooks/index.ts
```diff
@@ -0,0 +1,19 @@
...
+export const createIncidentQueryKey = (id: string) => ["getIncident", id];
+
+export const useIncidentQuery = (id = "") => {
+  return useQuery(createIncidentQueryKey(id), () =>
+    getIncident(id).then((response) => response.data)
+  );
+};
+
+export const useComponentsQuery = () => {
+  return useQuery(["allcomponents"], getTopologyComponents, {
```
moshloop, line 16: I don't think we want to use a topology cache as the only source for id lookups, we should have a seperate cache of just id to -> name, icon, type with very long expiration e.g. 2h 

the logic should be:

check local name cache, 
       check full topology cache (with 5m expiry)
            get /component/<id>

Prashanth-Nelli:
closing this PR as these changes are covered in https://github.com/flanksource/flanksource-ui/pull/553

PR #550, opened 2022-11-01
fix: persist restore sort type and direction

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli there are lots of errors as I navigate, the referer id should only be added if the component is not a direct child,

moshloop:
This does not seem to working correctly: https://watch.screencastify.com/v/mzZhXuwoIQ95AT8URQG2

Prashanth-Nelli:
@moshloop could you please review this. I have done the basic level of testing from my side after fixing the issues you have pointed out.

PR #554, opened 2022-11-03
feat: add grouping to config list

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli When sorting by type column, can we sort by both type and name ?

moshloop:
@Prashanth-Nelli and also remove the `[0]` and `-` cell values

moshloop:
@Prashanth-Nelli  Still doesn't look right

Screenshot 2022-11-08 at 11 55 10

Issue #556, opened 2022-11-06
/configs SidebarLayout component does not update the state for passed down props on window.innerWidth change (fix included)


tompston:
`SidebarLayout` component in `src/components/Layout/SidebarLayout.tsx` uses `window.innerWidth` value to toggle the state of the navigation. As the value is not reactive, it seems to update only once, when the component is loaded. 

`SidebarLayout` also has a `SideNav` component that needs the correct `collapseSidebar` prop to toggle the elements correctly, based on the current width of the window

<img width="253" alt="Screenshot 2022-11-06 at 21 22 41" src="https://user-images.githubusercontent.com/82293948/200190810-dae0208f-d2c2-4c5d-80dd-894a55ebcd50.png">

### Fix

1. Create a custom hook that gets the current inner width and height on change.

`src/hooks/useWindowDimensions.ts`

```jsx
// useWindowDimensions.ts
import { useState, useEffect } from 'react';

/**
 * Straight up copy from the lower link, with changed naming for more informative variables
 * https://stackoverflow.com/questions/36862334/get-viewport-window-height-in-reactjs
 * 
 * @example usage inside the component to get the latest value
 * const { windowHeight, windowWidth } = useWindowDimensions();
 * 
 */
function getWindowDimensions() {
    const { innerWidth: windowWidth, innerHeight: windowHeight } = window;
    return {
        windowWidth,
        windowHeight
    };
}

export default function useWindowDimensions() {
    const [windowDimensions, setWindowDimensions] = useState(getWindowDimensions());

    useEffect(() => {
        function handleResize() {
            setWindowDimensions(getWindowDimensions());
        }

        window.addEventListener('resize', handleResize);
        return () => window.removeEventListener('resize', handleResize);
    }, []);

    return windowDimensions;
}
```

2. Import and use the hook in the `src/components/Layout/SidebarLayout.tsx` component

```jsx
// new custom hook to get the value of the window dims
import useWindowDimensions from "../../hooks/useWindowDimensions"

export function SidebarLayout({ navigation, settingsNav }: Props) {
  ...
  // add this block inside the component to update the state for the collapseSidebar hook value

  /**
   * On window.innerWidth change,
   *  if window.innerWidth < 1024, set the toggle sidebar to be true
   *  else show the full sidebar
   */
  useEffect(() => {
    if (windowWidth < 1024) {
      setCollapseSidebar(true);
    } else {
      setCollapseSidebar(false);
    }
  }, [windowWidth])
}

PR #557, opened 2022-11-06
fixed issue 556

tompston:
Notes about the fixed problem can be found here https://github.com/flanksource/flanksource-ui/issues/556


moshloop:
@tompston I still see that issue if I resize the window down until the menu collapses, and then try and expand it

PR #558, opened 2022-11-07
536 logs improvements

### LINKED ISSUE #536: Logs - Improvements
- [x] Add labels column
- [x] Shrink timestamp column, set max-width of labels column and message to grow to remaining space
- [ ] Add filter option in components selection to show deleted components (default to off), 
- [x] Switch to time range component
- [ ] Only show components alive during time range

mainawycliffe:
Closes #536


moshloop:
@mainawycliffe need to add an empty search results

mainawycliffe:
@moshloop you mean where there are no query params - i.e. all logs?

mainawycliffe:
@moshloop Yeah, let me make the changes

moshloop:
@mainawycliffe When expanding the list of tags, the column widths change - I am expecting them to remain the same and only that single row height changes

mainawycliffe:
@moshloop spent a while playing with this, and two possible solutions I came across is to show one item per row and having a fixed width for the for the column or a combination of both. This is because the width of the label is unknown to begin with and makes it hard to predict how this will behave for different logs.

moshloop:
@mainawycliffe - We always want a fixed width for the columns (maybe dependant on size of screen)

mainawycliffe:
@moshloop I have set the width to fixed for all columns, with timestamp fitting content while message and labels having equal width and also made it possible to show only one column of labels with expand button as shown below:

image
image
image

moshloop:
@mainawycliffe This doesn't match v8 example: https://tanstack.com/table/v8/docs/examples/react/column-sizing ?

mainawycliffe:
@moshloop still working on it, was unable to continue last night when the API went down

mainawycliffe:
@moshloop is this the kind of resizing you had in mind? Currently, doesn't save to local storage, I am finalizing this right now.

moshloop:
@mainawycliffe This is still not working like the example:

* changing column width should not impact the width of other columns* 
* it should start out equal width it needs a more sensible default
* the resize indicator is too much - it is probaby more suited to the hover style

mainawycliffe:
@moshloop In order to not affect other column widths, we would the width of the table to be fluid, i.e., not 100 percent; at the moment, the column re-adjusts based on the available width (100%), but if it can grow to over 100% and shrink below 100%, then this is possible. We would start with a sensible width and then move on from there. Is this the desired behavior?

moshloop:
@mainawycliffe The table width should be full-width of the screen, and start with sensible widths for the columns which can then be changed - Something like this:  https://tanstack.com/table/v8/docs/examples/react/full-width-resizable-table is

mainawycliffe:
@moshloop that's what I went with
image

moshloop:
@mainawycliffe trying to make the labels smaller, just increases the width of the table and makes it scrollable

mainawycliffe:
@moshloop I have made the table width a little larger; now it shows scrollbars, and reduced the labels column.

moshloop:
@mainawycliffe this is still not working right - The timestamp column is not resizable but still seems to change width based on the other columns and easily gets to be shorter than the text

mainawycliffe:
@moshloop when the width of the table is fixed to 100%, the width of the timestamp column is resizing, despite giving it constraints for max size and min size and setting the column not autoresize, the column still autosizes. Trying to figure out a solution for this.

mainawycliffe:
@moshloop  Is this the kind of column resize indicator you had in mind? 
image

src/api/services/topology.ts
```diff
@@ -49,7 +50,14 @@ export const getCanaries = async (params) => {
...
 
 export const getTopologyComponents = () => {
-  return IncidentCommander.get(`/component_names`);
+  return IncidentCommander.get<TopologyComponentItem[]>(`/component_names`);
+};
+
+export const getTopologyComponentByID = async (topologyID: string) => {
+  const res = await CanaryChecker.get<TopologyComponentItem[]>(
+    `/db/component_names?id=eq.${topologyID}`
+  );
```
moshloop, line 59:
```suggestion
  const res = await IncidentCommander.get<TopologyComponentItem[]>(
    `/component_names?id=eq.${topologyID}`
  );
```

mainawycliffe: So, we should be using `IncidentCommander` instead for all DB requests?

moshloop:
@mainawycliffe  Can we rebase / squash - there are too many commits to merge, we shouldn't have more than 3 - 4

mainawycliffe:
@moshloop I have squashed the commits as much as I can, and now I have about 5 commits, In newer PRs, I should get better results by squashing regularly when everything is flesh in my mind.

PR #583, opened 2022-11-15
chore: code refactor of config fetch api to use config views

### LINKED ISSUE #501: Filters - use config_names, check_names rather than the underlying tables

Prashanth-Nelli:
closes #501


src/pages/config/config-details-changes.jsx
```diff
@@ -28,7 +28,7 @@ export function ConfigDetailsChangesPage() {
         setIsLoading(false);
       });
 
-    getConfig(id)
+    getConfigName(id)
```
moshloop, line 31: I think we should extract this into a ConfigBreadcrumb component + add the config icon (with a tooltip for type)

PR #587, opened 2022-11-17
fix: health page left side filters column margin issue fix

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli  I am expecting the sidebar to be fixed - it should only scroll if there are many filters, and it should scroll on its own. The body itself also shouldn't scroll, only the rows.

moshloop:
@Prashanth-Nelli this sidebar is fixed, but the table itself should only scroll the rows

mainawycliffe (approved):
Looks Good To Me

PR #589, opened 2022-11-17
fix: topology type evidence payload issue

Prashanth-Nelli:


src/components/TopologyCard/TopologyDropdownMenu.tsx
```diff
@@ -62,6 +62,7 @@ export const TopologyDropdownMenu = ({ topology }: IProps) => {
         onClose={() => setAttachAsAsset(false)}
         type={EvidenceType.Topology}
         evidence={{ id: topology.id }}
```
moshloop, line 64:
```suggestion
```

PR #593, opened 2022-11-21
feat: add group by tag support

Prashanth-Nelli:


moshloop:
Can we make the styling look more like the group by  on the health page? - The group by summary row should also only include the tag value, it seems to include a random selection of tag names and values ? Each row should then still have all the tags available

Prashanth-Nelli:
> Can we make the styling look more like the group by on the health page? - The group by summary row should also only include the tag value, it seems to include a random selection of tag names and values ? Each row should then still have all the tags available

Sure. I'll look into this.

PR #596, opened 2022-11-22
feat: add user list page view

Prashanth-Nelli:


src/App.tsx
```diff
@@ -96,6 +99,10 @@ export function IncidentManagerRoutes({ sidebar }: { sidebar: ReactNode }) {
         <Route index element={<HealthPage url={CANARY_API} />} />
       </Route>
 
+      <Route path="users" element={sidebar}>
```
moshloop, line 102: Can we move this under settings

src/components/UserList/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { formatLongDate } from "../../utils/date";
+import { DataTable } from "../DataTable";
+
+type UserListProps = {
+  data: any[];
+  isLoading?: boolean;
+};
+
+const DateCell = ({ cell: { value } }: { cell: { value: Date } }) =>
+  formatLongDate(value);
```
moshloop, line 12:
```suggestion
  formatDate(value);
```

src/pages/UsersPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+        return {
+          id: item.id,
+          created_at: new Date(item.created_at),
+          state: item.state,
+          state_changed_at: new Date(item.state_changed_at),
+          updated_at: new Date(item.updated_at),
+          name: `${item.traits?.name?.first ?? ""} ${
+            item.traits?.name?.last ?? ""
+          }`,
+          email: item.traits.email
```
moshloop, line 26: Shouldn't this be done in getRegisteredUsers() ?

Prashanth-Nelli: it depends, if there is a chance of using the same format data in multiple places then we can move it the getRegisteredUsers.

src/App.tsx
Prashanth-Nelli: sure

src/components/UserList/index.tsx
Prashanth-Nelli: We don't have an exported function formatDate. what date format are you expecting here.

src/pages/UsersPage.tsx
moshloop: It isn't so much whether it will be used in multiple places, but a separation of data / server handling and UI.

src/components/UserList/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  },
+  {
+    Header: "State",
+    accessor: "state",
+    Aggregated: ""
+  }
+];
+
+export function UserList({ data, isLoading }: UserListProps) {
+  const [queryParams, setQueryParams] = useSearchParams({
```
moshloop, line 39: Is this not something that should go into DataTable seeing as we do it in many places ?

src/components/UserList/index.tsx
moshloop: A date consistent with what we are using elsewhere - The long format shouldn't be used anywhere

Prashanth-Nelli: sure

src/components/Logs/index.tsx
```diff
@@ -7,31 +7,8 @@ type LogsProps = {
 };
 
 export function LogsViewer({ logs }: LogsProps) {
-  const [variant, setVariant] =
+  const [variant] =
```
mainawycliffe, line 10: **Something to note:** 

I have come across something like this before, and it's weird because if the variable doesn't change, then it should just be a constant rather than a state variable.

mainawycliffe (commented):
The user's table looks a little off; I think it's missing a border on top otherwise looks good


image
image

src/components/Logs/index.tsx
Prashanth-Nelli: sure, it make sense in cases where you want to memoize a value.

PR #598, opened 2022-11-22
fix: login, registration & recovery form style changes

Prashanth-Nelli:


pages/login.tsx
```diff
@@ -107,6 +107,9 @@ const Login: NextPage = () => {
               <Icon
                 name="flanksource-icon"
```
moshloop, line 108:
```suggestion
                name="flanksource"
```

moshloop: Can we switch to the flanksource logo and not the icon

Prashanth-Nelli: should I use this logo

Logo

PR #599, opened 2022-11-22
478-add-remote-dropdowns

### LINKED ISSUE #478: Add remote dropdowns
With icons:

* Config type -  `/db/config_types` (switch on config list page) 
* Analysis type /db/analysis_types`
* Change type -  `/db/changes types`

mainawycliffe:
Closes #478


src/components/ConfigsListFilters/ConfigListFilterByGroupByDropdown.tsx
```diff
@@ -0,0 +1,66 @@
+import { ComponentProps, useState } from "react";
+import { ReactSelectDropdown } from "../ReactSelectDropdown";
+
+type ConfigGroupByDropdownProps = Omit<
+  ComponentProps<typeof ReactSelectDropdown>,
+  "name"
+>;
+
+export default function ConfigListFilterByGroupByDropdown({
```
moshloop, line 9: Can we move these to a filters sub folder and then just call them GroupByDropdown, ConfigTypeDropdown etc.

moshloop:
@mainawycliffe please also squash changes were possible  each PR should ideally be a single commmit

src/api/services/configs.ts
```diff
@@ -134,3 +134,34 @@ export const getRelatedConfigs = async (configID: string) => {
   );
   return res.data;
 };
+
+export type ConfigTypeItem = {
+  config_type: string;
+};
+
+export const getConfigsTypes = async () => {
+  const res = await IncidentCommander.get<ConfigTypeItem[]>(`/config_types`);
```
Prashanth-Nelli, line 143: in case backend api errors data will be we need to handle those scenarios here.

mainawycliffe: @Prashanth-Nelli  throw an error in this case? Something like i`f(!data) throw error?`

Prashanth-Nelli: As we are using this in react queries, I think we need to handle cases where we are not getting the data. If it is already handled then we are good.

mainawycliffe: Just updating the types to reflect the reality that the data can be null.

Prashanth-Nelli (approved):

PR #605, opened 2022-11-28
feat: add invite user support

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli inviting users should be from the users list page

moshloop:
@Prashanth-Nelli  has this been linked to the backend api ?

Prashanth-Nelli:
> @Prashanth-Nelli has this been linked to the backend api ?

@moshloop yes but invite api is failing with mail sending error. Yash is looking into it.

src/components/InviteUserForm/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+                pattern: {
+                  value: /\S+@\S+\.\S+/,
+                  message: "Entered value does not match email format"
+                }
+              })}
+              className="h-full shadow-sm focus:ring-blue-500 focus:border-blue-500 py-2 sm:text-sm border-gray-300 rounded-md w-full"
+              type="email"
+            />
+          </div>
+          {errors.email && (
```
mainawycliffe, line 114: Maybe also add an error message for when the field is empty - required state

src/pages/UsersPage.tsx
```diff
@@ -1,14 +1,39 @@
...
 
 export function UsersPage() {
   const [users, setUsers] = useState<RegisteredUser[]>([]);
+  const [isOpen, setIsOpen] = useState(false);
   const { loading, setLoading } = useLoader();
 
+  const onSubmit = async (val: InviteUserFormValue) => {
+    try {
+      const { error } = await inviteUser(val);
+      if (!error) {
```
mainawycliffe, line 26: Just a little observation here, are we ignoring the error here on purpose? If that's not the case, we could re-throw it so that it can be caught by the catch, as shown below.

```ts
if (error) {
  throw error;
}

const userName = `${val.firstName} ${val.lastName}`;
toastSuccess(`${userName} invited successfully`);
setIsOpen(false);
fetchUsersList();
```

mainawycliffe (commented):
Just some light observations; otherwise looks good.

src/components/InviteUserForm/index.tsx
Prashanth-Nelli: We are using the same error message for both the cases, when value is not present & when the provided value is invalid.

mainawycliffe: Cool

PR #606, opened 2022-11-29
chore: update severity items

yashmehrotra:


src/components/IncidentSeverityTag/index.tsx
```diff
@@ -5,11 +5,18 @@ import {
 } from "react-icons/hi";
 import { IncidentSeverity } from "../../api/services/incident";
 
-const getSeverity = (severity: IncidentSeverity) => {
+const getSeverity = (severity: IncidentSeverity | string) => {
   switch (severity) {
-    case 2:
+    case "4":
```
moshloop, line 10: I don't understand why we have these numbers ?

Prashanth-Nelli: incident severity saved in the db is a number. we have this number to severity mapping because of that.

Prashanth-Nelli: Okay, I'll make changes from frontend and we need to write a db script or have work around in the front end to support both values.

PR #613, opened 2022-12-01
fix: use string version of severity instead of numeric

mainawycliffe:
The API is returning text version of severity - High, Low, Medium and this update severity to use those values directly.


src/components/IncidentDetails/AddResponder.tsx
```diff
@@ -281,14 +280,14 @@ export const AddResponder = ({
   });
   const watchAllFields = watch();
   const incidentDetails = useMemo(() => {
+    console.log(incident);
```
moshloop, line 283:
```suggestion
```

PR #622, opened 2022-12-07
feat: incident workflow

Prashanth-Nelli:


Prashanth-Nelli:
> @Prashanth-Nelli Can we make it look more like this: Screenshot 2022-12-14 at 09 19 52
> 
> Also ensure that we create reusable components for the select and for the evidence selection.

@moshloop I have made the changes as suggested. could you please review these changes.

mainawycliffe (approved):
LGTM

PR #625, opened 2022-12-07
Fix-incident-severity-bug-on-incident-create

mainawycliffe:


moshloop:
@mainawycliffe can you check the priority/severity combo on incident details it is rendering empty ?

mainawycliffe:
@moshloop cool

mainawycliffe:
@moshloop should be good now, it was another source of number severity and was the reason the issue was occuring

moshloop:
@mainawycliffe The dropdown now matches the full list of severities, but it still shows as empty

mainawycliffe:
@moshloop are we talking about this part here?
image

PR #631, opened 2022-12-12
fix: logs evidence table columns width issue fix

Prashanth-Nelli:


mainawycliffe (commented):
@Prashanth-Nelli The width of the time column is supposed to remain fixed, but it's not in the logs page. Maybe we should pass in the className and size of the time column as props to the LogsTable; this way, we can set the width based on where the table is being displayed.

mainawycliffe (approved):
LGTM

src/components/Hypothesis/EvidenceSection/index.tsx
```diff
@@ -19,10 +19,23 @@ import { LogsTable } from "../../Logs/Table/LogsTable";
 import { Icon } from "../../Icon";
 import { Button } from "../../Button";
 
+const ColumnSizes = {
+  Time: {
+    size: 64,
+    maxSize: 64
```
moshloop, line 25:
```suggestion
    size: 70,
    maxSize: 70
```

PR #633, opened 2022-12-13
Sidebar refactoring

moshloop:


src/components/Sidebars/configs.tsx
```diff
@@ -0,0 +1,90 @@
...
+  configId?: string;
+};
+
+export function ConfigsList({ topologyId, configId }: Props) {
+  const [configs, setConfigs] = useState<ConfigItem[]>([]);
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    async function fetchConfigs() {
```
mainawycliffe, line 21: Just something I have been meaning to bring up, we probably should start of switching all API calls to React Query; the User Experience will greatly improve due to its fine-grained state management and caching, making the data appear much faster.

mainawycliffe (approved):
LGTM, the comment on react-query is something that I have been thinking lately

mainawycliffe:
@moshloop should we start to generally avoid `index.tsx`, not a huge fan as it adds extra mental load when navigating in the UI.

PR #638, opened 2022-12-14
feat: extract reusable components out of incident filters bar component

### LINKED ISSUE #632: IncidentsFilterBar - Switch to common components
Need an IncidentType and IncidentType Dropdown comoonent that use React select dropdown, the filter bar should then just pass through classes to remove the border / etc..

Prashanth-Nelli:
closes #632


mainawycliffe (approved):
LGTM

PR #659, opened 2022-12-19
chore: move show hidden components toggle to preferences & improve sort dropdown styling

Prashanth-Nelli:


mainawycliffe (commented):
Looks good, but would be nice if we could center this icon vertically 
image

Prashanth-Nelli:
> Looks good, but would be nice if we could center this icon vertically image

Done

mainawycliffe (approved):
LGTM

PR #660, opened 2022-12-19
fix: add default behaviour for config list sorting & grouping

Prashanth-Nelli:


src/components/ConfigList/index.tsx
```diff
@@ -433,13 +433,22 @@ export interface Props {
...
 
-  const groupByField = queryParams.get("groupBy");
+  const groupByField = queryParams.get("groupBy") || "config_type";
+
+  useEffect(() => {
+    setQueryParams({
+      ...Object.fromEntries(queryParams),
+      groupBy: "config_type",
+      sortBy: "config_type",
+      sortOrder: "inc"
```
mainawycliffe, line 449: Shouldn't this be `asc` or `desc`?

mainawycliffe (commented):
Just a single question in my mind upon testing

mainawycliffe (approved):

PR #661, opened 2022-12-19
Back button on config detail requires multiple clicks #641

### LINKED ISSUE #641: Back button on config detail requires multiple clicks

mainawycliffe:
Closes #641


src/components/DataTable/index.tsx
```diff
@@ -49,41 +49,15 @@ export function DataTable<TableColumns, Data extends TableColumns>({
...
-  const sortBy = useMemo(() => {
-    const data = sortField
-      ? [
-          {
-            id: sortField,
-            desc: isSortOrderDesc
-          }
-        ]
-      : [];
-    if (sortField === "config_type" && usageSection === "config-list") {
```
Prashanth-Nelli, line 78: We need this to sort the config list by config_type & name whenever user tries sorting by config_type column. I see this logic is being removed in the refactor. could you please handle this.

mainawycliffe: I am lifting this to ConfigList, and then it passed it in via as the DataTable shouldn't care about specific implementation details. Any consumer of DataTable should now pass in SortBy, which is compatible with react-table, ensuring we reduce the no. of implementation details in the DataTable component. One question though. Do we want to sort by config type and then name?

src/pages/config/ConfigDetailsPage.tsx
```diff
@@ -1,40 +1,39 @@
...
-      })
-      .catch((err) => toastError(err))
-      .finally(() => {
-        setIsLoading(false);
-      });
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [id]);
+  const [checked, setChecked] = useState<Record<string, any>>({});
+
+  const { isLoading, data: configDetails } = useQuery(
```
Prashanth-Nelli, line 21: Please move this logic to react queries file where we have all the queries.

src/pages/config/ConfigList.tsx
```diff
@@ -30,9 +24,9 @@ export function ConfigListPage() {
   });
 
   const search = params.get("search");
-  const tag = decodeURIComponent(params.get("tag") || "All");
-  const configType = decodeURIComponent(params.get("type") || "All");
-  const groupByProp = decodeURIComponent(params.get("groupByProp") ?? "");
+  const tag = params.get("tag") || "All";
```
Prashanth-Nelli, line 27: decodeURIComponent is needed as some of the values has special characters which gets encoded when we set query params. so when we read we need to decode.

src/components/DataTable/index.tsx
Prashanth-Nelli: Moving this to ConfigList is fine. But we need to handle sorting by confg_type + name whenever user tries sorting by config_type

mainawycliffe: In that case, we can expose setSort or something that updates the sort state whenever it changes.

src/pages/config/ConfigList.tsx
mainawycliffe: Ah, makes sense.

src/pages/config/ConfigDetailsPage.tsx
mainawycliffe: The one disadvantage of doing this is we lose type inference and autocompletion on options, which is very in cases like [select](https://tkdodo.eu/blog/react-query-data-transformations), which is useful when you want to transform the data without affecting the cached data. I am still trying to figure out a way of wrapping around useQuery, without losing it.

src/hooks/usePartialUpdateSearchParams.tsx
```diff
@@ -0,0 +1,57 @@
...
+): [URLSearchParams, (newParams: Record<string, string | string[]>) => void] {
+  const [params, setParams] = useSearchParams(init);
+
+  const setPartialParams = useCallback(
+    (newParams: Record<string, string | string[]>) => {
+      let paramsChanged = false;
+      Object.entries(newParams).forEach(([key, value]) => {
+        console.log(key, value);
+        // remove the key if the value is falsy
+        if (!value) {
```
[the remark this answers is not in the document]
mainawycliffe: What do you mean?

src/components/DataTable/index.tsx
Prashanth-Nelli: I think moving lines from 78 to 85 here https://github.com/flanksource/flanksource-ui/pull/661/files#diff-fe40532367ba1cce6d73ad3ef76bfdbdbfebc7fd4b865afc5f704ebf3689c3b2R459 will handle the use case.

src/hooks/usePartialUpdateSearchParams.tsx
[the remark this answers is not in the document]
Prashanth-Nelli: we can only conclude whether there is a change or not when we compare newParams value & current params value. here we are not doing that. Here we are only checking whether the value is empty in new params.

src/pages/config/ConfigDetailsPage.tsx
Prashanth-Nelli: Okay, but this will lead to code duplication. we can solve type inference problem later.

src/hooks/usePartialUpdateSearchParams.tsx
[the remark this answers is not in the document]
mainawycliffe: This is fixed no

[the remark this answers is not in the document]
mainawycliffe: Should be re-usable by any other component with a custom sort implementation.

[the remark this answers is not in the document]
mainawycliffe: In this case, if the new value is set to empty, it's safe to assume the user wants to delete it. If the new value is set to something else, that's when we compare; the logic flows after that.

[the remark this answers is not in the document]
mainawycliffe: This is how the logic flows:

```ts
 // remove the key if the value is falsy
if (!value) {
  params.delete(key);
  paramsChanged = true;
  return;
}
// if the value hasn't changed, don't update the URL
if (Array.isArray(value)) {
  // loop through the new values and see if any of them are different
  if (params.getAll(key).filter((x) => !value.includes(x)).length > 0) {
    value.forEach((v) => {
      paramsChanged = true;
      params.append(key, v);
    });
  }
} else {
  if (params.get(key) !== value) {
    paramsChanged = true;
    params.set(key, value);
  }
}
```

Prashanth-Nelli (approved):
LGTM

PR #669, opened 2022-12-20
656-component-dropdown-should-show-none-instead-of-all

### LINKED ISSUE #656: In Logs Table, Component dropdown should show none instead of all
Since we don't show all logs but rather one per component, the default entry should switch to none instead of all

image

DesmondSanctity:
Closes #656 

image

### DIFF
--- a/src/components/FilterLogs/FilterLogsByComponent.tsx
+++ b/src/components/FilterLogs/FilterLogsByComponent.tsx
@@ -6,9 +6,9 @@ import { Icon } from "../Icon";
 import { ReactSelectDropdown, StateOption } from "../ReactSelectDropdown";
 
 export const defaultSelections = {
-  all: {
-    description: "All",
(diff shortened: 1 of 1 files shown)


mainawycliffe:
One more thing, we probably need to show the value none when none is selected instead of an empty input 
image

DesmondSanctity:
> One more thing, we probably need to show the value none when none is selected instead of an empty input image

Okay. I will check that.

DesmondSanctity:
I have fixed that.

image

mainawycliffe:
@DesmondSanctity please rebase against the latest changes on the main branch

DesmondSanctity:
okay

mainawycliffe (approved):

mainawycliffe:
@DesmondSanctity Thanks for the PR

PR #673, opened 2022-12-21
feat: config analysis description improvements

Prashanth-Nelli:


src/components/ConfigInsights/index.tsx
```diff
@@ -39,6 +49,13 @@ function ConfigInsightsDetails({ configID }: Props) {
         `/api/db/config_analysis?config_id=eq.${configID}`
       );
       const data = (await res.json()) as ConfigTypeInsights[];
+      data.forEach((item) => {
```
mainawycliffe, line 52: Just nitpicking here; we should avoid mutations like this, as they can introduce bugs that are very unpredictable and very hard to track (Javascript passes objects by reference, so you may be mutating something being used somewhere else). Instead, use the .map array function, which is the same number of lines but produces a new object.

src/utils/common.js
```diff
@@ -32,3 +33,26 @@ export function toQueryString(paramsObject) {
     )
     .join("&");
 }
+
+export const sanitizeHTMLContent = (htmlString) => {
```
mainawycliffe, line 37: Can we please add types for these input parameters? We will turn on typescript strictness in the future, and we need to make this a habit.

src/utils/common.js
```diff
@@ -32,3 +33,26 @@ export function toQueryString(paramsObject) {
...
+  return sanitizeHtml(htmlString, {
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat(["img"]),
+    allowedAttributes: {
+      "*": ["style", "class"]
+    }
+  });
+};
+
+export const sanitizeHTMLContentToText = (htmlString) => {
```
mainawycliffe, line 46: Same for this as well

src/utils/common.js
```diff
@@ -32,3 +33,26 @@ export function toQueryString(paramsObject) {
...
+  const html = sanitizeHTMLContent(htmlString);
+  const div = document.createElement("div");
+  div.innerHTML = html;
+  return div.textContent;
+};
+
+export const truncateText = (txt = "", limit) => {
```
mainawycliffe, line 53: And this as well.

mainawycliffe (commented):
Just a few points to note otherwise looks good.

mainawycliffe (approved):

moshloop:
@Prashanth-Nelli  - we are already using dompurify ?

src/utils/common.ts
```diff
@@ -0,0 +1,58 @@
+import { parse } from "qs";
+import sanitizeHtml from "sanitize-html";
+
+export function toFixedIfNecessary(value: string, dp: number) {
+  return +parseFloat(value).toFixed(dp);
+}
+
+export function capitalizeFirstLetter(string: string) {
```
moshloop, line 8: Isn't this a css transform ?

src/utils/common.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  return `${parseFloat((bytes / k ** i).toFixed(dm))} ${sizes[i]}`;
+}
+
+export function searchParamsToObj(searchParams: Record<string, string>) {
+  return parse(searchParams.toString());
+}
+
+export function toQueryString(paramsObject: Record<string, string>) {
+  return Object.keys(paramsObject)
```
moshloop, line 29: Should we use `qs` for this ?

src/utils/common.ts
Prashanth-Nelli: we have a use case where we capitalise a specific words in a sentence. css transform doesn't work in that case.

Prashanth-Nelli:
> @Prashanth-Nelli - we are already using dompurify ?

I'll switch to this module and remove sanitize-html dependency.

PR #680, opened 2022-12-26
feat: add support for linking config analysis as evidence to incident

### LINKED ISSUE #651: Attach analysis as evidence

Prashanth-Nelli:
Closes #651

### DIFF
--- a/src/components/ConfigAnalysisLink/ConfigAnalysisLink.tsx
+++ b/src/components/ConfigAnalysisLink/ConfigAnalysisLink.tsx
@@ -1,0 +1,168 @@
+import { useState, useEffect } from "react";
+import ReactTooltip from "react-tooltip";
+import { EvidenceType } from "../../api/services/evidence";
+import { formatISODate, formatLongDate } from "../../utils/date";
+import { AttachEvidenceDialog } from "../AttachEvidenceDialog";
+import { ConfigTypeInsights } from "../ConfigInsights";
+import ConfigInsightsIcon from "../ConfigInsightsIcon";
+import { Description } from "../Description/description";
+import { Icon, Avatar } from "../Icon";
+import { Modal } from "../Modal";
+
+type Props = {
+  configAnalysis: ConfigTypeInsights;
+  viewType?: "summary" | "detailed";
+} & React.HTMLProps<HTMLDivElement>;
+
+export function ConfigAnalysisLink({
+  configAnalysis,
+  viewType = "summary",
+  ...rest
+}: Props) {
+  const [attachEvidence, setAttachEvidence] = useState(false);
+  const [open, setOpen] = useState(false);
+
+  useEffect(() => {
+    ReactTooltip.rebuild();
+  });
+
+  return (
+    <div
+      className="flex flex-col cursor-pointer  space-y-1"
+      {...rest}
+      onClick={() => {
+        setOpen(true);
+      }}
+    >
+      <Modal
+        title={
+          <>
+            <ConfigInsightsIcon analysis={configAnalysis} />
+            {configAnalysis.analyzer}
+          </>
+        }
+        open={open}
+        onClose={(e) => {
+          e?.stopPropagation();
+          setOpen(false);
+        }}
+        size="large"
+        bodyClass=""
+      >
+        {configAnalysis?.id && (
+          <>
+            <div className="flex flex-col px-4 py-4 text-gray-500 space-y-2">
+              <div className="flex flex-col space-y-6">
+                <div className="grid grid-cols-1 gap-x-4 gap-y-8 sm:grid-cols-4">
+                  <div className="sm:col-span-1">
+                    <Description
+                      label="Type"
+                      value={
+                        <>
+                          <Icon
+                            name={configAnalysis?.analysis_type}
+                            secondary="diff"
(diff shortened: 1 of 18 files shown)


moshloop:
+ We should create an AnalysisLink component that covers the icon / name / age and popup and then reuse that that for the evidence body
+ The modal title should be icon and analysis name

moshloop:
+ apply the same approach for changes
+ In the popup don't use age, use date and the standard timestamp format

Prashanth-Nelli:
@moshloop made changes as suggested

moshloop:
@Prashanth-Nelli - The dialog title should be ec2-no-public-ip, not security which is indicated in the icon

src/components/ConfigAnalysisLink/ConfigAnalysisLink.tsx
```diff
@@ -0,0 +1,168 @@
...
+        }}
+        size="large"
+        bodyClass=""
+      >
+        {configAnalysis?.id && (
+          <>
+            <div className="flex flex-col px-4 py-4 text-gray-500 space-y-2">
+              <div className="flex flex-col space-y-6">
+                <div className="grid grid-cols-1 gap-x-4 gap-y-8 sm:grid-cols-4">
+                  <div className="sm:col-span-1">
```
moshloop, line 57: This should be refactored into DescriptionCard with labelStyle="top", there shouldn't be any styling at this level it should all be in DescriptionCard - this ensures that all descriptions are styled consistently throughout - e.g. Config analysis and changes descriptions now have different fonts

moshloop:
- Don't show null details / changes
- The resource attached to change / analysis should be prefixed in the title (and be a link)
e.g. {AWS Icon} ACC123 / {security icon} vpc-flow-logs-enabled
- The type should have the type icon, not a generic diff icon
- Don't use NA, just have it blank
- It appears to show the current date, The dates (and labels) for a config analysis should be First Observed and Last Observed
- The "identified @ / changes made" should be removed in evidence and definition of done display - The dates for these should only display in table format for the Changes / Analysis sidebars. 
- The config change analysis evidence should should just be {resource icon} {resource name} / {change icon} {change name}
- The link component should never render anything more than a link (to a config) or to a popup, the date should be the loop i.e. don't remove the table in ConfigInsights / Analysis

src/components/ConfigDetailsChanges/ConfigDetailsChanges.tsx
```diff
@@ -39,85 +76,136 @@ export function ConfigDetailsChanges({
...
+        open={open}
+        onClose={(e) => {
+          e?.stopPropagation();
+          setOpen(false);
+        }}
+        size="large"
+        bodyClass=""
+      >
+        <div
+          className="flex flex-col h-full"
```
moshloop, line 105: Why do we need this styling? should't it be part of the modal ?

src/components/DescriptionCard/index.tsx
```diff
@@ -1,28 +1,56 @@
-type DescriptionCardProps = {
+import clsx from "clsx";
+
+type DescriptionCardProps = React.HTMLProps<HTMLDivElement> & {
   items: {
-    label?: string;
+    label?: string | React.ReactNode;
     value: React.ReactNode;
   }[];
+  noOfCols?: number;
```
moshloop, line 8:
```suggestion
  columns?: number;
```

src/components/ConfigDetailsChanges/ConfigDetailsChanges.tsx
```diff
@@ -39,85 +76,136 @@ export function ConfigDetailsChanges({
...
+            <Icon
+              name={changeDetails?.change_type}
+              secondary="diff"
+              className="w-5 h-auto pr-1"
+            />
+            {changeDetails?.change_type}
+          </>
+        }
+        open={open}
+        onClose={(e) => {
```
moshloop, line 97: Why do we need this ?

moshloop (changes requested):

src/components/ConfigDetailsChanges/ConfigDetailsChanges.tsx
Prashanth-Nelli: child elements layout can column wise or row wise and it depends what kind of layout we want to show inside Modal. so we can't just move this into Modal. 

I think we can assign these classes to `bodyClass` prop of `Modal`.

src/components/ConfigDetailsChanges/ConfigDetailsChanges.tsx
Prashanth-Nelli: When a Modal is opened on top of another Modal close is not working properly. In order to fix that issue I have added this.

moshloop: But then shouldn't it be on all modals ? and/or add a comment on why it is needed

Prashanth-Nelli: I'll add a comment

moshloop:
@Prashanth-Nelli 
- The config name and icon should not be shown in the config page - it is already in the title.
- Elsewhere it is missing the config_icon

PR #681, opened 2022-12-27
fix: populate creator details properly

Prashanth-Nelli:


src/api/schemaResources.ts
```diff
@@ -3,6 +3,7 @@ import {
   SchemaBackends
 } from "../components/SchemaResourcePage/resourceTypes";
 import { CanaryCheckerDB, ConfigDB, IncidentCommander } from "./axios";
+import { AVATAR_INFO } from "./services/incident";
```
moshloop, line 6: I don't think AVATAR_INFO is specific to incidents - maybe it should go in another file ?

PR #691, opened 2022-12-28
676-sorting-time-ranges-should-not-be-pushed-to-history

### LINKED ISSUE #676: Sorting / Time ranges should not be pushed to history

mainawycliffe:
- [x] Topology page 
- [x] Health Page

Closes #676


moshloop:
@mainawycliffe - I think we still want to update the URL, just not push that URL to history

PR #702, opened 2022-12-30
feat: add config list remote sort filter support

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli The search seems to be broken now ?

Prashanth-Nelli:
@moshloop search issue is fixed.

PR #721, opened 2023-01-12
711-topology-add-snapshot-action

### LINKED ISSUE #711: Topology - Add snapshot action
See https://github.com/flanksource/incident-commander/blob/main/cmd/server.go#L63

Popup a dialog with the options

- Date Range
- Include Related Components

mainawycliffe:
NB: It's timing out on related components because of Lambdas 3 sec timeout.

Closes #711

### DIFF
--- a/src/api/axios.js
+++ b/src/api/axios.js
@@ -60,6 +60,16 @@ export const Config = axios.create({
   }
 });
 
+export const Snapshot = axios.create({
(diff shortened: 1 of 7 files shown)


mainawycliffe:

mainawycliffe:
@moshloop this is good to go

PR #729, opened 2023-01-13
feat: add frontend & backend version info in the profile dropdown

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli the frontend version info should come from package.json

src/api/query-hooks/index.ts
```diff
@@ -23,14 +23,36 @@ import {
...
+        data.data = {
+          ...versionInfo,
+          frontend: packageJson.version,
+          backend: versionInfo.Version
+        };
+        return data;
+      });
+    },
+    {
+      staleTime: defaultStaleTime,
```
mainawycliffe, line 50: Just a nitpick -> I have enabled this globally for all queries https://github.com/flanksource/flanksource-ui/blob/d3f6fb083e72914b46203830f4cc0cabf9ef96ec/src/App.tsx#L52.

src/api/query-hooks/index.ts
```diff
@@ -23,14 +23,36 @@ import {
...
+      enabled: true
```
mainawycliffe, line 51: Another nitpick, shouldn't this be true by default?

src/components/VersionInfo/VersionInfo.tsx
```diff
@@ -0,0 +1,13 @@
+import { useVersionInfo } from "../../api/query-hooks";
+
+export function VersionInfo() {
+  const { data } = useVersionInfo();
+  const versionInfo = data?.data as any;
```
mainawycliffe, line 5: We should avoid casting as any and instead provide the necessary types.

mainawycliffe (approved):
LGTM

PR #737, opened 2023-01-17
fix: page data refresh on hypothesis status is changed to disproven

### LINKED ISSUE #352: Page doesn't refresh after disproving hypothesis in delete window

Prashanth-Nelli:
Closes #352


src/components/mutations/useUpdateHypothesisMutation.js
```diff
@@ -1,61 +1,19 @@
...
-          )
-        );
-      }
+    onSuccess: () => {
+      queryClient.refetchQueries(incidentQueryKey);
     },
-    onError: (err, newData, context) => {
-      queryClient.setQueryData(incidentQueryKey, context.previousIncident);
-      queryClient.invalidateQueries(createIncidentQueryKey(incidentId));
+    onError: (err) => {
```
mainawycliffe, line 15: Doesn't Axios already handle these APIs errors or only for GET?

mainawycliffe (approved):
LGTM

src/components/mutations/useUpdateHypothesisMutation.js
Prashanth-Nelli: Yeah, we already have a global error handler added in one of the response interceptor. This is unnecessary here, I'll remove this.

PR #757, opened 2023-01-19
fix: login page improvements

Prashanth-Nelli:


mainawycliffe (approved):
Are e removing the option to create account and using invites instead?

moshloop:
@mainawycliffe  - Correct it should be invite only

PR #758, opened 2023-01-19
fix: attach evidence modal improvements

Prashanth-Nelli:


mainawycliffe (approved):
lgtm

moshloop:
@Prashanth-Nelli 

- When attaching to an incident:
-  If there is single hypothesis, auto-select the first hypotheis 
- Make the hypothesis required

moshloop:
@Prashanth-Nelli Only when attaching should hypothesis be required, otherwise it should default (either in the background) or in the UI

PR #765, opened 2023-01-19
fix: do not reloading icons when refetching dod

### LINKED ISSUE #761: Definition of done list - don't show reloading icon
The loading state should only be show on first load, thereafter it should not be shown as it causes the UI to flicker

mainawycliffe:
Fixes #761


Prashanth-Nelli (approved):
LGTM

PR #768, opened 2023-01-24
feat: extract cost info panel & integrate into config details side bar section

### LINKED ISSUE #717: Topology Details - Add cost panel
+ Create reusable shared components for all cost-side panels

Prashanth-Nelli:
Closes #717

### DIFF
--- a/src/components/CostDetails/CostDetails.tsx
+++ b/src/components/CostDetails/CostDetails.tsx
@@ -1,0 +1,122 @@
+import { useMemo } from "react";
+import { FaDollarSign } from "react-icons/fa";
+import Title from "../Title/title";
+import CollapsiblePanel from "../CollapsiblePanel";
+import { Loading } from "../Loading";
+
+export type CostsData = {
+  cost_per_minute?: number;
+  cost_total_1d?: number;
+  cost_total_7d?: number;
+  cost_total_30d?: number;
+};
+
+type CostDetailsTableProps = CostsData;
+
+type CostInfoPanelProps = CostsData & {
+  loading?: boolean;
+  title: string;
+};
+
+type FormatCurrencyProps = {
+  value: number | string | undefined;
+  defaultValue?: string | null | React.ReactNode | number;
+  hideMinimumValue?: boolean;
+};
+
+type CostInfoProps = {
+  label: string;
+  value: number | string | undefined;
+  defaultValue?: string | null | React.ReactNode | number;
+};
+
+function isCostsDataEmpty({
+  cost_per_minute,
+  cost_total_1d,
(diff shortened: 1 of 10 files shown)


moshloop:
@Prashanth-Nelli extracting the cost component is the first step, but it also needs to be added into the topology details which also have a cost associated.

src/components/TopologySidebar/TopologyCost.tsx
```diff
@@ -0,0 +1,29 @@
+import { useMemo } from "react";
+import { Topology } from "../../context/TopologyPageContext";
+import {
+  arrangeCostDataForInfoPanel,
+  CostInfoPanel,
+  isCostsDataEmpty
```
moshloop, line 6: Why do we need the arrangeCost / isCosts empty functions - can't those be called from inside the CostInfoPanel

Prashanth-Nelli: These are exposed because of following reasons.
- we don't want to hard code the view behaviour into the CostInfoPanel in case of data absence.
- cost data format can be different so we want to covert that into the format ConstInfoPanel understands.
   - we provided one implementation for our existing cost data format.
   
I don't have strong opinions around this. If we get the cost data in same format across all api's and we don't have use case for having different ui behaviour in case of cost data absence. I can move this into CostInfoPanel component.

@moshloop please let me know your thoughts.

moshloop: > we don't want to hard code the view behaviour into the CostInfoPanel i 

Why not, if the data is empty it doesn't render anything, or renders an empty state if specified

> cost data format can be different

It shouldn't be

src/components/CostDetails/CostDetails.tsx
```diff
@@ -0,0 +1,122 @@
...
+}: CostsData) {
+  return (
+    cost_per_minute === undefined &&
+    cost_total_1d === undefined &&
+    cost_total_7d === undefined &&
+    cost_total_30d === undefined
+  );
+}
+
+export function FormatCurrency({
```
mainawycliffe, line 47: Don't we already have something like this?

mainawycliffe (commented):
Just something small to think about, do we want to hide a cost column when empty or show nothing underneath?

image

Otherwise, looks good.

image

Prashanth-Nelli:
> Just something small to think about, do we want to hide a cost column when empty or show nothing underneath?
> 
> image
> 
> Otherwise, looks good.

in case of specific cost details absence we are keeping it empty in the existing functionality.

mainawycliffe (approved):
LGTM

PR #771, opened 2023-01-25
fix: incident card styling fixes

### LINKED ISSUE #763: Incident types rendering issues.
The incident items should look the same as sub-components re garding padding/icon sizing placement etc.. - We should have a StatusComponent that handles this rendering. 
Screenshot 2023-01-19 at 11 43 40

Prashanth-Nelli:
Closes #763

### DIFF
--- a/src/components/IncidentStatusTag/index.tsx
+++ b/src/components/IncidentStatusTag/index.tsx
@@ -1,6 +1,7 @@
 import clsx from "clsx";
 import { capitalize } from "lodash";
 import { IncidentStatus } from "../../api/services/incident";
+import { Tag } from "../Tag/Tag";
 
 interface IProps {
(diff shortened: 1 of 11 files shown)


src/components/IncidentStatusTag/index.tsx
```diff
@@ -15,19 +15,17 @@ const statusClsMap = {
 
 export function IncidentStatusTag({ status, size = "md", className }: IProps) {
   return (
-    <button
-      type="button"
+    <div
```
moshloop, line 18: Rather than applying the same style in both places, can we not create a new component with just the styling that is re-used ?

Prashanth-Nelli: sure

PR #784, opened 2023-01-31
feat: add support for action plan & comments view in incident details page

### LINKED ISSUE #780: Incident Detail - Comment View
Create a separate view in another tab that shows a comment-focused view (i.e. no hypothesis shown anywhere) - the 2 tabs would be called "Action Plan", "Comments"

Prashanth-Nelli:
Closes #780 

### Review feedback ###

- [x] The comment bar should always be fixed at the bottom - this should behave exactly like slack
- [x] The tab and body have different borders - Can we not extract the tab styling on the health page into a Tabs and Tab component?
- [x] The react select-dropdown width is too small, and it should popup above, not below the comment bar


moshloop:
@Prashanth-Nelli - I should be able to add new comments in the comment view, and when adding the comment, be able to select the hypothesis the comment is for

moshloop:
The evidence items should also be shown - i.e. it is the action plan, minus the hypothesis grouping

moshloop:
@Prashanth-Nelli  The comment box should look the same and be on a single line,  and be at the bottom, sorting of comments should be disabled - it is always oldest to newest, scrolled to the bottom of the page

The hypothesis selection should default to the root,

Prashanth-Nelli:
@moshloop there can be multiple root hypothesises. so the select dropdown will have one of the root hypothesis pre selected

moshloop:
@Prashanth-Nelli 

- The comment bar, and card bar at the top should be fixed, only the previous comments should scroll
- The date should also go next to Username, not below, also the commented/added evidence part can come out
- Can we embed the hypothesis tag into the input,  and then click on it to show the dropdown above

moshloop:
@Prashanth-Nelli - the tag should be right aligned, not left aligned - The dropdown should be react select popout - the comment bar should not move

moshloop:
@Prashanth-Nelli 

- The comment bar should always be fixed at the bottom - this should behave exactly like slack 
- The tab and body have different borders - Can we not extract the tab styling on the health page into a Tabs and Tab component?
- The react select-dropdown width is too small, and it should popup above, not below the comment bar

moshloop:
@Prashanth-Nelli  the comment bar is still not pinned to the bottom, on incidents with few comments it is in the middle of the screen and for those with many comments, it is off the screen and unusable

PR #791, opened 2023-02-02
781-add-action-bar-in-topology-sidebar

### LINKED ISSUE #781: Add action bar in topology sidebar
i.e.all actions in the burger menu should be actions at the top of the of the sidebar.

mainawycliffe:
Closes #781

### DIFF
--- a/src/components/TopologySidebar/TopologyActionBar.tsx
+++ b/src/components/TopologySidebar/TopologyActionBar.tsx
@@ -1,0 +1,272 @@
+import React, { useState } from "react";
+import { IconType } from "react-icons";
+import { BiShow, BiHide, BiZoomIn } from "react-icons/bi";
+import { ImTree } from "react-icons/im";
+import { MdAlarmAdd, MdTableRows } from "react-icons/md";
+import { useNavigate } from "react-router-dom";
+import useUpdateComponentMutation from "../../api/query-hooks/mutations/useUpdateComponentMutation";
+import { EvidenceType } from "../../api/services/evidence";
+import { Topology } from "../../context/TopologyPageContext";
+import { AttachEvidenceDialog } from "../AttachEvidenceDialog";
+import { Button } from "../Button";
+import TopologySnapshotModal from "../TopologyCard/TopologySnapshotModal";
+
+type TopologyActionItem = {
+  label: string;
+  icon: IconType;
+  isShown: (
+    topology: Topology,
+    displayPoint: "TopologyCard" | "TopologySidebar"
+  ) => boolean;
+  /**
+   *
+   * A headless component that wraps the child component and provides the onClick handler
+   * and icon and text props.
+   *
+   * The onclick handler is unique to each action item, and the icon and text
+   * and hence the need to create them in the container component and pass them
+   * to child component.
+   *
+   */
+  ContainerComponent: React.FC<{
+    child: React.FC<{
+      onClick: () => void;
+      icon: React.ReactNode;
+      text: React.ReactNode;
+    }>;
+    topology: Topology;
+    onRefresh?: () => void;
+    icon: React.ReactNode;
+    text: React.ReactNode;
+    /**
+     *
+     * A callback that is called to open a modal such as the attach evidence
+     * dialog or the topology snapshot modal.
+     *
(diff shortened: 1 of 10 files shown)


moshloop:
- Can make the burger menu and the sidebar actions both use the same actions object so that changes to actions only need to take place in  1 place
- Can we use the normal button component?  it is not very clear that these are buttons

moshloop:
@mainawycliffe - I don't see any changes?  This must use the https://github.com/flanksource/flanksource-ui/tree/main/src/components/Button component, If we make styling changes to that component, they must propogate to every button.

mainawycliffe:
@moshloop Just doing some cleanup, will be pushing the changes in the next few minutes

mainawycliffe:
@moshloop I have made the changes, what do you think?

src/components/TopologySidebar/TopologyActionBar.tsx
```diff
@@ -0,0 +1,286 @@
...
+  onClick = () => {},
+  icon,
+  text
+}: TopologyActionLinkProps) {
+  return (
+    <div className="flex w-full flex-col items-center p-1 h-full">
+      <Button
+        icon={icon}
+        text={text}
+        className="flex w-full h-full flex-col space-y-2 justify-center items-center btn-primary"
```
moshloop, line 217: We should not be changing the class on the buttons - the only thing that should every change is the color. The style should be normal button icon on the left, text on the right - rectangle, not square

src/components/TopologySidebar/TopologyActionBar.tsx
```diff
@@ -0,0 +1,286 @@
...
+      const navigate = useNavigate();
+
+      return (
+        <ChildComponent
+          onClick={() => navigate(`/topology/${topology.id}`)}
+          icon={icon}
+          text={text}
+        />
+      );
+    }
```
moshloop, line 109: Why can't this just be onClick: {() => navigate(...)}

src/components/TopologySidebar/TopologyActionBar.tsx
```diff
@@ -0,0 +1,286 @@
...
+  label: string;
+  /**
+   *
+   * The icon to be shown in the action item, this is a react component and can
+   * be rendered directly
+   *
+   * e.g. <Icon {...props} />
+   *
+   */
+  Icon: IconType;
```
moshloop, line 27:
```suggestion
  icon: IconType;
```

src/components/TopologySidebar/TopologyActionBar.tsx
```diff
@@ -0,0 +1,286 @@
...
+import { EvidenceType } from "../../api/services/evidence";
+import { Topology } from "../../context/TopologyPageContext";
+import { AttachEvidenceDialog } from "../AttachEvidenceDialog";
+import { Button } from "../Button";
+import TopologySnapshotModal from "../TopologyCard/TopologySnapshotModal";
+
+type TopologyActionItem = {
+  /**
+   * The label to be shown in the action item
+   */
```
moshloop, line 17: Don't add comments for the obvious

src/components/TopologySidebar/TopologyActionBar.tsx
mainawycliffe: Due to how react works, components must start with a capital letter. Hence the options are either this or alias every time we need to use them in order to get intellisense and be recognized as components.

src/components/TopologySidebar/TopologyActionBar.tsx
mainawycliffe: The logic to determine URL to navigate to would also need to be extracted away (properties also don't work as they don't have access to the topology object needed to determine the component details), which would affect readability and ease of use when consuming the list and also adding new action items.

moshloop:
@mainawycliffe can we remove the panel border, and switch to white buttons like on the incident page.

mainawycliffe:
@moshloop you mean like this

image

moshloop:
@mainawycliffe I don't see any buttons now ?

mainawycliffe:
@moshloop sorry, messed the rebase, fixed it now.

PR #793, opened 2023-02-03
786- Job History Improvements

### LINKED ISSUE #786: Job History improvements
- In the error popup use a table style, not list style
- Use the actual date and time rather than time ago
- Duration - convert to human duration
- Add filters on job type, resource type, id, success
- Move sorting server side
- Convert from CamelCase to Camel Case in the job names

### LINKED ISSUE #1138: Close out 793
https://github.com/flanksource/flanksource-ui/pull/793

DesmondSanctity:
Fixes #786  and Fixes #1138

- [x] In the error popup use a table style, not list style
- [x] Use the actual date and time rather than time ago
- [x] Duration - convert to human duration
- [x] Add filters on job type, resource type, id, success
- [x] Move sorting server side
- [x] Convert from CamelCase to Camel Case in the job names

### DIFF
--- a/src/utils/date.ts
+++ b/src/utils/date.ts
@@ -1,6 +1,7 @@
 import dayjs, { ConfigType as DateConfig, ManipulateType } from "dayjs";
 import isBetween from "dayjs/plugin/isBetween";
 import isToday from "dayjs/plugin/isToday";
+import duration from "dayjs/plugin/duration";
 import isSameOrAfter from "dayjs/plugin/isSameOrAfter";
 import relativeTime from "dayjs/plugin/relativeTime";
 import updateLocale from "dayjs/plugin/updateLocale";
@@ -15,6 +16,7 @@ dayjs.extend(isSameOrAfter);
 dayjs.extend(relativeTime);
 dayjs.extend(updateLocale);
 dayjs.extend(utc);
+dayjs.extend(duration);
 
 dayjs.updateLocale("en", {
   relativeTime: {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -300,6 +300,7 @@ function SidebarWrapper() {
 
   const pathTrack = [
     "users",
+    "jobs",
     "teams",
     "incident_rules",
     "config_scrapers",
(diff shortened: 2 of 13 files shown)


moshloop:
@DesmondSanctity can you finish up the remaining items ?  If you have a look at the configs or incidents list page it should have the patterns that you can use.

DesmondSanctity:
Yeah. I'm almost done with filters. The server side sorting how do I know the endpoints?

moshloop:
@DesmondSanctity it is the same endpoint, just additional params that need to be added - check how it works for configs and incidents

DesmondSanctity:
Okay got it.

DesmondSanctity:
@moshloop This is done. Please check and drop feedbacks.

moshloop:
@DesmondSanctity 
- I think there is an issue with duration - it might be assuming seconds instead of millisconds - we also probably want to use formatDuruation from utils to get values like '1h' or '1m23s'
- There is no filter for status
- The popup window should not have s/n - it should be just a single column
- The sort, should also be integrated into the table - see the config table

moshloop:
@DesmondSanctity please only import date functions from utils.js, not from dayjs directly - The duration is wrong, most jobs return in milliseconds not seconds and it should display 25ms, or 1s or 2m1s

DesmondSanctity:
> @DesmondSanctity please only import date functions from utils.js, not from dayjs directly - The duration is wrong, most jobs return in milliseconds not seconds and it should display 25ms, or 1s or 2m1s

Noted. I'll change it.

DesmondSanctity:
@moshloop you can check now. Thanks

DesmondSanctity:
@moshloop any other improvements for Job history??

mainawycliffe:
@moshloop Do we want to create views for Job History Names and Resource IDs, the way we are currently fetching them will be hit by DB limits as the job history grows.

src/utils/date.ts
```diff
@@ -116,6 +118,29 @@ export const formatTimeRange = (date: string | Date) => {
   });
 };
 
+export const formatDateToSeconds = (date: string | Date) =>
```
moshloop, line 121: Don't use a new function/format

for the duration.Humanize()

+ Customize durations across the board:

Use d vs day,

mainawycliffe: @moshloop is this what you had in mind?

mainawycliffe:
@moshloop this should be good to go now

PR #794, opened 2023-02-03
feat: allow add comments as definition of done

### LINKED ISSUE #719: Manual Definition of Done
- Provide the ability to add a comment as a definition of done (either inline, or in the wizard)
- Allow comments to be manually resolved

mainawycliffe:
Closed #719

### DIFF
--- a/src/api/query-hooks/mutations/evidence.tsx
+++ b/src/api/query-hooks/mutations/evidence.tsx
@@ -1,5 +1,9 @@
 import { useMutation } from "@tanstack/react-query";
-import { Evidence, updateEvidence } from "../../services/evidence";
+import {
+  createEvidence,
+  Evidence,
+  updateEvidence
+} from "../../services/evidence";
 
 /**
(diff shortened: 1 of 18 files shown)


moshloop:
@mainawycliffe 
- It looks like it creates a new hypothesis? 
- The styling on doesn't look very consistent - The 3 options should look the same, it think it might be better to have a 1st page wizard which lets you select the type, and then the next page which one (or in the case of manual, just the comment) - The comment should be limited to 50 chars otherwise the rendering will be bad
- The icon should be different, maybe a human to indicate a human needs to perform it?
- There should be an option to mark as done / unmark as done

moshloop:
@mainawycliffe  Can we change it from a textarea to a text input ? And also make the tick green ?

moshloop:
Just a plain input - in v2.0 we might assign these to responders but not for this release

mainawycliffe:
@moshloop done

PR #799, opened 2023-02-06
fix: clicking on incident type opens incident listing page with type preselected

### LINKED ISSUE #797: Topology card - incident links don't work
Clicking on the incident type or severity should open the incidents page with the filters pre-selected

Prashanth-Nelli:
Closes #797

### DIFF
--- a/src/api/services/incident.ts
+++ b/src/api/services/incident.ts
@@ -49,7 +49,7 @@ export interface Incident extends NewIncident {
 }
 
 export const searchIncident = (query: string) => {
-  const hypotheses = `hypotheses!hypotheses_incident_id_fkey(id, type)`;
+  const hypotheses = `hypotheses(id, type)`;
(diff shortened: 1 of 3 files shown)


moshloop:
@Prashanth-Nelli the component should be preselected - but it looks like the query is not working ?

Prashanth-Nelli:
@moshloop currently we are just using root topology item id as component id in the link. 
Ideally we shouldn't be including component id in the url, because we can multiple incidents with different component id's under a topology.

Should I remove the component id from the link ?.

moshloop:
@Prashanth-Nelli - If I click on a component card I want to see the incidents for that card so the id must be in the URL, and the incident list filter must only show incidents with that component as an attached evidence item

Prashanth-Nelli:
@moshloop the issue is due to wrongly formed query, I have fixed the query filtering is working fine now.

Issue #800, opened 2023-02-09
Collapsible panels - Make the caret icon target larger


moshloop:
Clicking on the title doesn't work and you need to click on the caret, the click target should expand to the full width up until the filters

Prashanth-Nelli:
Handled in the PR https://github.com/flanksource/flanksource-ui/pull/802

PR #802, opened 2023-02-10
fix: make sidebar panels occupy same height and expand to full height

### LINKED ISSUE #801: Sidebar panels should fill height equally
i.e. Each panel should be an equal height, and should stretch to the full height of the page.

Prashanth-Nelli:
Closes #801 & #800


moshloop:
@Prashanth-Nelli 

- Icons on the incident sidebar are not correct
- The title is not clickable with action items - I think the actions should be left, and then the remaining right space is clickable
- The Cost panel should not be equal height as it will always have whitespace, maybe we should integrate it into details ?
- The icon sizes are not consistent

Prashanth-Nelli:
@moshloop I can modify the sidebar container to customise height allocation on the basis of configuration at the child level.

example:-

* for fixed height: `data-panel-height="100px"`
* for occupying same height as other panels: `data-panel-height="auto"`

This attribute will be specified at the collapsible panel level.

Prashanth-Nelli:
> @Prashanth-Nelli
> 
> * Icons on the incident sidebar are not correct
> * The title is not clickable with action items - I think the actions should be left, and then the remaining right space is clickable
> * The Cost panel should not be equal height as it will always have whitespace, maybe we should integrate it into details ?
> * The icon sizes are not consistent

Fixex

PR #816, opened 2023-02-20
fix: while editing resource spec hide remaining tabs

Prashanth-Nelli:
This is done to keep user in editing flow once he clicks on edit instead of allowing him to navigate to other sections.


mainawycliffe (approved):
LGTM

PR #821, opened 2023-02-22
fix: retain scroll & expansion state of config list table

Prashanth-Nelli:
- [x] remove tooltips
- [x] style the popover it should look different from background ui & only one popover should be shown.
- [x] only clicking on more should open popover
- [x] if there is one item don't show popover
- [x] auto close popover after 5 seconds after opened.
- [x] improve performance of config list

### DIFF
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -1,5 +1,5 @@
 import clsx from "clsx";
-import { LegacyRef } from "react";
+import { LegacyRef, useCallback, useEffect } from "react";
 import { FaCog } from "react-icons/fa";
 import { useOnMouseActivity } from "../../hooks/useMouseActivity";
 import { ClickableSvg } from "../ClickableSvg/ClickableSvg";
@@ -8,6 +8,9 @@ type PopoverProps = {
   popoverIcon?: React.ReactNode;
   children: React.ReactNode;
   placement?: "right" | "left";
+  toggle?: React.ReactNode;
+  autoCloseTimeInMS?: number;
+  menuClass?: string;
 } & React.HTMLProps<HTMLDivElement>;
 
 export default function Popover({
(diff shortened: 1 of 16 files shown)


moshloop:
@Prashanth-Nelli can you also have a look at https://www.loom.com/share/40e8a3c1e37c4497bb0683d056e91444 - I think this was happening also before this PR

moshloop:
@Prashanth-Nelli 

- We should still show at least 1 analysis, with horizontal ... for more - Same applies to tags, Show until it overflows and then 3 dots
- Use a popover, not a full modal 
- For changes - just count them

moshloop:
@Prashanth-Nelli  - It is not clear that there are more items or that the tag is clickable - might need something like "3 more"

Prashanth-Nelli:
Sure

src/components/Popover/Popover.tsx
```diff
@@ -27,23 +29,27 @@ export default function Popover({
...
+        </button>
+      )}
+      {toggle}
       <div
         role="menu"
         aria-orientation="vertical"
         aria-labelledby="menu-button"
         className={clsx(
-          "flex flex-col origin-top-right absolute mt-5 w-96 z-50 divide-y divide-gray-100 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none capitalize",
+          "flex top-8 flex-col origin-top-right absolute w-96 z-50 divide-y divide-gray-100 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none capitalize",
```
moshloop, line 52:
```suggestion
          "flex top-8 flex-col origin-top-right absolute w-96 z-50 divide-y divide-gray-100 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none",
```

moshloop:
@Prashanth-Nelli - the click target on the +4 more is too small - it should probably be the entire cell

PR #828, opened 2023-02-27
fix: eagerly update query client level incident state to render posted comment instantly

### LINKED ISSUE #813: Add comment to list before sending to server (and remove if server event fails)

Prashanth-Nelli:
Closes #813 

- [x] refactor incident details data fetching and integrate redux store
- [x] space should be at the top like slack, recent message should be shown just above the comment box.
- [x] re order tabs to comments and action plan
- [x] add a title to hypothesis select control
- [x] remove commented from the comments title.
- [x] hypothesis selection is getting lost after commenting.


mainawycliffe (approved):
LGTM

Just one minor nitpick, make we should start switching these old mutations to the useMutation hook from react-query, then the cache update can be done within the hook but totally up to you, as this works as well.

moshloop:
I am not sure this is working - I am expecting an the comment to be added instantly after enter, the comment box to clear, and the I can then start typing again and hit enter

Prashanth-Nelli:
@moshloop I have only covered the below requirements
-  I am expecting an the comment to be added instantly after enter

I'll look work on the remaining requirements.

Prashanth-Nelli:
> LGTM
> 
> Just one minor nitpick, make we should start switching these old mutations to the useMutation hook from react-query, then the cache update can be done within the hook but totally up to you, as this works as well.

Made changes as suggested

moshloop:
@Prashanth-Nelli the comment is added instantly, but then it is hidden and reshown

PR #832, opened 2023-02-28
feat: switch tag list source to config_tags view not configs

### LINKED ISSUE #814: Config tag list - Sort + style label

mainawycliffe:
Closes #814


moshloop:
@mainawycliffe 
- Can we reduce the font-size in the dropdown + increase the max-width there are many values that are cutoff
- Can we style the label in a gray font

mainawycliffe:
@moshloop cool, let me do that

mainawycliffe:
@moshloop something like this?
image
image

PR #849, opened 2023-03-02
fix: avoid making logs api call unneccessarily when component is not selected

Prashanth-Nelli:


src/api/query-hooks/index.ts
```diff
@@ -386,3 +387,39 @@ export function useConfigAnalysisQuery(
     options
   );
 }
+
+export function useComponentGetLogsQuery(
```
mainawycliffe, line 391: The file is getting way too big; we should start putting new and related mutations in new files, named appropriately.

mainawycliffe (approved):
LGTM, just some thoughts for the future.

src/api/query-hooks/index.ts
Prashanth-Nelli: I already have a plan for this. I am going to integrate state management library as part of config list improvements issue. I'll be refactoring this into separate files in that PR.

Thanks for bringing it up.

PR #850, opened 2023-03-02
fix: show parent link in topology sidebar detail panel

Prashanth-Nelli:


src/components/TopologyDetails/index.tsx
```diff
@@ -23,11 +23,7 @@ export default function TopologyDetails({ topology, refererId }: Props) {
 
   var items = [];
 
-  if (
-    refererId != null &&
-    topology.parent_id != null &&
-    topology.parent_id !== refererId
-  ) {
+  if (topology.parent_id != null) {
```
moshloop, line 26: I don't think we want to show parent  always - only when navigating soft links and the parent in the breadcrumb is different to the hard parent

Prashanth-Nelli:
Closing this as the issue is already handled.

PR #856, opened 2023-03-03
742-Disable actions on incident details page once incident is closed

### LINKED ISSUE #742: disable actions on the incident details page once the incident is closed.
- modifying definition of done
- adding a new factor
- deleting (hypothesis, evidence, comments etc)
- hypothesis status changes

DesmondSanctity:
This  PR fixes #742

### DIFF
--- a/src/components/Dropdown/SubtleDropdown.tsx
+++ b/src/components/Dropdown/SubtleDropdown.tsx
@@ -4,28 +4,27 @@ import { CheckIcon } from "@heroicons/react/solid";
 import { Controller } from "react-hook-form";
 import clsx from "clsx";
 import { IItem } from "../../types/IItem";
-import { IncidentStatus } from "../../api/services/incident";
 
 interface Props {
   className: string;
+  disabled: boolean;
   value: string;
   label: string;
   control: any;
   onChange: (value: string) => void;
   name: string;
   items: { [k: string]: IItem };
-  checkStatus?: IncidentStatus;
 }
 
 export function SubtleDropdown({
   className,
+  disabled = false,
   label,
   control,
   items = {},
   name,
   onChange = () => {},
-  value,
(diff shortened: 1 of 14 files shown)


src/components/Dropdown/SubtleDropdown.tsx
```diff
@@ -4,6 +4,7 @@ import { CheckIcon } from "@heroicons/react/solid";
 import { Controller } from "react-hook-form";
 import clsx from "clsx";
 import { IItem } from "../../types/IItem";
+import { IncidentStatus } from "../../api/services/incident";
```
moshloop, line 7: SubtleDropdown is a generic component it shouldn't be referencing IncidentStatus - same applies for many of the other components - A generic component should not be aware of specific uses .

DesmondSanctity: > SubtleDropdown is a generic component it shouldn't be referencing IncidentStatus - same applies for many of the other components - A generic component should not be aware of specific uses . 

I'll fix that. Thanks for the review

src/components/Dropdown/SubtleDropdown.tsx
```diff
@@ -22,7 +23,8 @@ export function SubtleDropdown({
   items = {},
   name,
   onChange = () => {},
-  value
+  value,
+  checkStatus
```
moshloop, line 27: checkStatus should not be a field on SubtleDropdown - SubtleDropdown is a generic component it can be disabled/hidden or have different modes of operation but it should have any fields that are specific to 1 use case.

DesmondSanctity: @moshloop I used disabled instead of checkStatus in the latest commit. I think it can be applied to any other components using the Dropdown. Please do check. Thanks

PR #857, opened 2023-03-03
fix: fix topology item job item being empty when not

### LINKED ISSUE #846: Job History table for Topology / Scrapers / etc is empty

mainawycliffe:
Closes #846


moshloop:
@mainawycliffe  I don't think we need the resource, resource id or job type as they will always be the same

mainawycliffe:
@moshloop do we want to remove the two fields for all tables or just the ones for individual items - component job history or also the main table that contains everything?

Issue #790, opened 2023-02-02
UI - Remove drop shadow on pass/fail Health dots


halverson:
On the Health Screen, the red and green dots in the "Health" column should not have drop shadows as this makes them harder to read. Removing the drop-shadows will increase the contrast and make them more distinguishable on the white background.

![Screenshot 2023-02-02 at 11 53 34](https://user-images.githubusercontent.com/6161540/216292989-9b6731ac-7f31-43ae-beba-c0ea5c30d362.png)

gitstart:
@halverson I would like to pick this up.

PR #904, opened 2023-03-04
835-Improved mixed state pass/fail Health dots

### LINKED ISSUE #835: UI - Improved mixed state pass/fail Health dots
When there are mixed results of the pass/fail dots in the Health column on a top-level item, they appear as a red dot with a green border. This is difficult to read and may be almost impossible for some users.

Screenshot 2023-02-28 at 14 15 31

Change to some variation where there is a more clear divide between the two colours, such as a gap or white line between the red and green. Options attached in an image.

Mixed Dots Options

DesmondSanctity:
This PR fixes #835 & #790


moshloop:
We no longer use mixed state dots

PR #905, opened 2023-03-06
feat: extract details table component & integrated into insights, incidents & changes sections

### LINKED ISSUE #844: Extract DetailsTable from Insights / Incidents
Create a new DetailsTable for rendering insights, incidents, changes, etc

- Use insights as base
- Add a dashed border between rows
- reduce the y padding to 1 (except for the first row maintain at 2 top)

Prashanth-Nelli:
Closes #844


src/components/ConfigInsights/index.tsx
```diff
@@ -39,24 +38,54 @@ type Props = {
...
+      })
+      .map((item) => {
+        return {
+          age: relativeDateTime(item.first_observed),
+          analysis: (
+            <div
+              key={item.id}
+              data-html={true}
+              data-tip={item.sanitizedMessageTxt}
+              data-class="max-w-[20rem]"
```
moshloop, line 82: Why do we need this ?

Prashanth-Nelli: Config analysis string can be html with tags to render it properly in the tooltip we need these attributes.

PR #914, opened 2023-03-09
feat: integrate check summary api & code refactor

Prashanth-Nelli:
- [x] assign fixed width to last transition time column and rename it to last transition.
- [x] there shouldn't be any ui flickering.
- [x] style last transition time column value similar to latency & uptime columns.
- [x] fix popup height issue on bigger screens.
- [x] there is mismatch between status shown and the uptime value verify it.
- [x] if the last runtime is greater than 15mins show relative time after health icon.
- [x] for mixed state use orange colour dot instead of the current one.
- [x] don't send start query param to summary api.
- [x] in the health check popup there is a big gap between the icon & title.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -86,7 +86,7 @@ const settingsNav = {
 
 export type SettingsNavigationItems = typeof settingsNav;
 
-const CANARY_API = "/api/canary/api";
+const CANARY_API = "/api/canary/api/summary";
 
 export function HealthRoutes({ sidebar }: { sidebar: ReactNode }) {
   return (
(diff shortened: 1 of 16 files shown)


moshloop:
@Prashanth-Nelli it looks like there is any extra space between `6h `and `ago`
- The logic for showing aging checks doesn't look like it is working

Screenshot 2023-03-10 at 16 47 23

moshloop:
@Prashanth-Nelli the hide passing filter appears to be broken now, also fix the width of the health and last transition columns

Prashanth-Nelli:
> @Prashanth-Nelli the hide passing filter appears to be broken now, also fix the width of the health and last transition columns

fixed.

moshloop:
@Prashanth-Nelli when expanding a group, the  name columns resize which creates a jarring effect

PR #1012, opened 2023-03-22
fix: check status history table ui/ux fixes

### LINKED ISSUE #919: Check Status Table fixes
- The table should be full height, with the pager at the bottom - i.e. the whiteshapce is between the last row and the pager, not between the pager and the footer
- It seems to be sending to many queries:
- 
Screenshot 2023-03-10 at 16 48 27

- The paging buttons don't appear to work

Prashanth-Nelli:
Closes #919


moshloop:
@Prashanth-Nelli The table does not seem to query based on the time range selected, nor is the paging done server side? It also seems like there is a duplicate request

PR #1019, opened 2023-03-24
fix: incident details sidebar ui/ux improvements

### LINKED ISSUE #812: Incident Details Side panels taking up space even when they don't need to
If this is intentional, may be add a vertically centered No details found or Collapse if they are empty.

image

Prashanth-Nelli:
Closes #812 & #966

- [x] Reduce the definition of done and responders panels to half the size of the others.
- [x] arrange incident details in this order:- type, status, priority, commanders, responders, start & duration.
- [x] show empty state block and align it centrally. 
- [x] We should resize all the panels if one of them is collapsed


moshloop:
@Prashanth-Nelli I don't see any change ?

PR #1029, opened 2023-03-27
937 config changes panel add virtual paging

### LINKED ISSUE #937: Config changes panel - Add virtual paging

Prashanth-Nelli:
Closes #937 

### Description
- added a new component `InfiniteTable` which provides virtual paging support.
- modified existing `DataTable` component to add support for virtual paging.
- code refactored to integrate above changes

### Test Plan
- open the [url](https://deploy-preview-1029--goofy-euclid-75956c.netlify.app/configs/0185578e-22c6-1165-1ef1-97f707f264ff)
- login with your credentials
- go to the config changes section in the sidebar
- try scrolling and observe data getting loaded as we scroll to the end.
- when all the data is loaded observe the total count in the config changes header and the sum of all the records returned in the api calls should match.

### Demo
demo

### DIFF
--- a/src/api/query-hooks/index.ts
+++ b/src/api/query-hooks/index.ts
@@ -6,7 +6,7 @@ import {
   getAllConfigsMatchingQuery,
   getConfig,
   getConfigAnalysis,
-  getConfigChange,
+  getConfigChanges,
   getConfigInsight,
   getConfigInsights,
   getConfigTagsList,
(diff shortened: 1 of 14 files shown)


Prashanth-Nelli:
> @Prashanth-Nelli Virtual paging wouldn't have a pager - See https://tanstack.com/table/v8/docs/examples/react/virtualized-rows
> 
> A page of size of 25 would be better,
> 
> It also appears to be doing a query per row ?

@moshloop  made changes as requested.

moshloop:
@Prashanth-Nelli this is not working - On configs with a small number of changes it just loops and continously sends requests to the server, on items with many changes it reenders fine and works for the first few scrolls and then it just gets stuck and keeps rendering the cells in a loop (consolo.log("diff icon not  found") just repeats at dozens of times per secon

moshloop:
- I still see 2 initial requests, and then 5 other queries in bulk with no offset. 
- 
Screenshot 2023-04-04 at 11 07 26

Lets switch to react-window and vanilla html table

moshloop:
@Prashanth-Nelli please extract the non-virtualized changes to another PR.

The paging is still not working, all it does is increase the page size from 50 -> 100 -> 200 -> 400 - The page size should remain constant and only the offset incremented by 50

Prashanth-Nelli:
@moshloop I have made changes as suggested.

moshloop:
@Prashanth-Nelli I still see a query for the full list in addition to the paging query on load

PR #1033, opened 2023-03-28
fix: fix scrollbars showing on dropdown in modals

### LINKED ISSUE #993: Attach Evidence Type Dropdown causing Scrollbars to appear
Ideally, this should be at the very top of every element, growing outside the modal if needed

image

mainawycliffe:
Closes #993


moshloop:
@mainawycliffe Shouldn't this fix be applied to the modal itself ?

PR #1062, opened 2023-04-05
fix: add attach to incident action button in config details page

### LINKED ISSUE #1054: Config Detail - Add Attach to incident action
We seem to have lost this

Prashanth-Nelli:
Closes #1054


moshloop:
@Prashanth-Nelli this button should be available even if a line is not selected - It should probably also go above the details/tags panel like topology and use the same button styling

PR #1065, opened 2023-04-05
fix: removed redundant fields in check details tab

### LINKED ISSUE #954: Check details tab - remove redundant fields
- name 
- type (add a tooltip to the icon)
- uptime
- status
-

Prashanth-Nelli:
Closes #954


moshloop:
@Prashanth-Nelli I think we remove the fields we want - name/type/status/uptime are all included in the header - we want only the other fields - Also the tab should fill the height

moshloop:
@Prashanth-Nelli - The initial loading message is shown at the very top of the div and not in the middle.

Prashanth-Nelli:
@moshloop I have fixed the issue.

PR #1076, opened 2023-04-12
fix: ui improvements

Prashanth-Nelli:
- add count badge component do display counts and to adjust style based on the value.
- topology sidebar menu items are re arranged.
- definition done item views ui/ux fixes.
- sorted definition of done items before displaying.
- auto collapse collapsible panel when there is no data based on the dataCount property.
- fixed the issue of comment box on comments view of incidents details is hidden when there are no hypothesis.
- fixed the styling issues of tab in active state.
- sliding sidebar storybook changes.
- fixed the issue of manually added definition of done item not being shown until refresh is clicked.

### DIFF
--- a/src/api/query-hooks/index.ts
+++ b/src/api/query-hooks/index.ts
@@ -43,7 +43,8 @@ export const useVersionInfo = () => {
 
 export const useIncidentQuery = (id: string) => {
   return useQuery(createIncidentQueryKey(id), () => getIncident(id), {
-    staleTime: defaultStaleTime
+    staleTime: defaultStaleTime,
+    enabled: !!id
   });
 };
 
--- a/src/api/query-hooks/index.ts
+++ b/src/api/query-hooks/index.ts
@@ -43,7 +43,6 @@ export const useVersionInfo = () => {
 
 export const useIncidentQuery = (id: string) => {
   return useQuery(createIncidentQueryKey(id), () => getIncident(id), {
-    staleTime: defaultStaleTime,
     enabled: !!id
   });
 };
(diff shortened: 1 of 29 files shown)


src/api/query-hooks/index.ts
```diff
@@ -43,7 +43,8 @@ export const useVersionInfo = () => {
 
 export const useIncidentQuery = (id: string) => {
   return useQuery(createIncidentQueryKey(id), () => getIncident(id), {
-    staleTime: defaultStaleTime
+    staleTime: defaultStaleTime,
```
mainawycliffe, line 46: Isn't this set globally for all queries, at the `QueryClientProvider` client?

src/api/query-hooks/mutations/evidence.tsx
```diff
@@ -12,9 +13,13 @@ import {
...
-export function useUpdateEvidenceMutation(param?: {
-  onSuccess?: (evidence: Evidence[]) => void;
-}) {
+export function useUpdateEvidenceMutation(
+  param: {
+    onSuccess?: (evidence: Evidence[]) => void;
+  },
+  incidentId: string
+) {
+  const { refetchIncident } = useIncidentState(incidentId);
```
mainawycliffe, line 22: May be a better approach for this is to use: `queryClient.refetchQueries([keys])` to refetch queries, after an update:

```ts
const queryClient = useQueryClient()

queryClient.refetchQueries(keys)
```

mainawycliffe (commented):
Just a couple of comments on some minor things, otherwise looks good.

mainawycliffe (approved):
LGTM

PR #1078, opened 2023-04-13
feat: add config analysis summary line to the topology card

Prashanth-Nelli:


moshloop:
@Prashanth-Nelli I think it will need a label called Insights and the badges should match the existing badges - it shouldn't look different to any other row.

Prashanth-Nelli:
@moshloop I have made changes as suggested. This is how it looks right now.

Screenshot 2023-04-13 at 6 35 06 PM

moshloop:
@Prashanth-Nelli I think we will need to move the colors i.e. the analysis type icon should be dark gray, and the numbers color based on severity (red for critical, orange for warning, gray for info)

Prashanth-Nelli:
@moshloop added popup view for displaying insights details on clicking on the insights line item.

Screenshot 2023-04-17 at 7 10 39 PM

PR #1100, opened 2023-04-27
feat: add support for enable & disabling features based on the configuration

### LINKED ISSUE #1082: Add environment variable to disable features
When specified 

- DISABLE_HEALTH 
- DISABLE_CONFIG
- DISABLE_TEAMS
- DISABLE_LOGS
- DISABLE_TOPOLOGY
- DISABLE_INCIDENTS

Prashanth-Nelli:
Closes #1082


src/App.tsx
```diff
@@ -52,13 +52,35 @@ import ErrorPage from "./components/Errors/ErrorPage";
 import { useQuery } from "@tanstack/react-query";
 import { AxiosError } from "axios";
 import { Provider } from "jotai";
+import { permissionService } from "./services/permissions/permissionsService";
+import { RESOURCES } from "./services/permissions/resources";
```
moshloop, line 56: Use CamelCase for variable names

src/App.tsx
```diff
@@ -71,7 +93,8 @@ const settingsNav = {
     {
       name: "Users",
       href: "/settings/users",
-      icon: HiUser
+      icon: HiUser,
+      resouceName: RESOURCES["SETTINGS.USERS"]
```
moshloop, line 97: Why isn't this a variable like the others ?

src/services/permissions/permissionsService.ts
```diff
@@ -0,0 +1,31 @@
...
+  properties: Property[] = [];
+
+  async loadProperties() {
+    const data = await getProperties();
+    this.properties = data.data || [];
+    return data;
+  }
+
+  isFeatureDisabled(resouceName: string) {
+    const name = `DISABLE_${resouceName}`;
```
moshloop, line 22:
```suggestion
    const name = `${resouceName.toLower()}.disable`;
```

src/App.tsx
Prashanth-Nelli: It is a constant so I went with all caps

src/App.tsx
Prashanth-Nelli: I have grouped all resources under `RESOURCES` constant so that while using we can just import `RESOURCES` instead of importing each of them individually.

src/services/permissions/resources.ts
```diff
@@ -0,0 +1,14 @@
+export const resouces = {
+  topology: "topology",
+  health: "health",
+  incidents: "incidents",
+  config: "config",
+  logs: "logs",
+  "settings.users": "settings.users",
+  "settings.teams": "settings.teams",
+  "settings.rules": "settings.rules",
+  "settings.config_scraper": "settings.config_scraper",
```
moshloop, line 10: The settings are linked to the primary property e.g. if config.disable == true hide both configs and config scraper tabs

PR #1103, opened 2023-04-28
feat: Signup Button Added in the SignIn Screen that routes to the Signup Page.

mprabs:
Added a signup button inside the login screen.

image


moshloop:
This is as designed

PR #1105, opened 2023-05-03
chore: update types and db calls to reflect db changes

mainawycliffe:


src/components/ConfigAnalysis/ConfigInsightsColumns.tsx
```diff
@@ -26,8 +26,8 @@ export const ConfigInsightsColumns: ColumnDef<
           >
             <Icon
               className="w-4 h-4"
-              name={config?.external_type}
-              secondary={config?.config_type}
+              name={config?.type}
+              secondary={config?.config_class}
             />
```
moshloop, line 31: Use this style all over instread of name={configItem.type || configItem.class}

src/components/ConfigItem/storybook-mock.ts
```diff
@@ -1,7 +1,7 @@
 export const MOCK_DATA = [
   {
     id: "0181ebeb-b9e7-009b-a28d-72c2e6d13f1b",
-    config_type: "JIRA",
+    config_class: "JIRA",
```
moshloop, line 4: these should all be type

src/components/HealthChecksSummary/index.tsx
```diff
@@ -28,12 +28,12 @@ export function HealthChecksSummary({
       warning: 0
     };
 
-    checks.forEach((check) => {
-      const status = check.status;
-      summary.healthy += status === "healthy" ? 1 : 0;
-      summary.warning += status === "warning" ? 1 : 0;
-      summary.unhealthy += status === "unhealthy" ? 1 : 0;
-    });
+    // checks.forEach((check) => {
```
moshloop, line 31: ?

mainawycliffe: checks are not an array, but an object, I am following up with Yash to understand what this means

src/api/services/configs.ts
```diff
@@ -82,13 +81,13 @@ export const getAllChanges = (
...
     if (queryParams[key] && queryParams[key] !== "All") {
-      const queryKey = key === "type" ? `config.config_type` : key;
+      const queryKey = key === "type" ? `config.type` : key;
       queryString += `&${queryKey}=eq.${queryParams[key]}`;
     }
   });
   return resolve(
     ConfigDB.get<ConfigTypeChanges[]>(
-      `/config_changes?order=created_at.desc${pagingParams}&select=*,config:config_names!inner(id,name,config_type)${queryString}`,
+      `/config_changes?order=created_at.desc${pagingParams}&select=*,config:config_names!inner(id,name,config_class)${queryString}`,
```
moshloop, line 90:
```suggestion
      `/config_changes?order=created_at.desc${pagingParams}&select=*,config:config_names!inner(id,name,type)${queryString}`,
```

src/api/services/configs.ts
```diff
@@ -180,15 +179,15 @@ export const searchConfigs = (type: string, input: string) => {
     : "";
   return resolve<ConfigItem[]>(
     ConfigDB.get(
-      `/configs?select=id,external_id,name,config_type,analysis,changes&config_type=ilike.${type}${orCondition}`
+      `/configs?select=id,external_id,name,type,analysis,changes&config_class=ilike.${type}${orCondition}`
```
moshloop, line 182:
```suggestion
      `/configs?select=id,external_id,name,type,analysis,changes&type=ilike.${type}${orCondition}`
```

src/api/services/configs.ts
mainawycliffe: It seems we need to update config_names as well to add type as well.

PR #1125, opened 2023-05-11
fix: fix health schedule field incorrect mapping

mainawycliffe:


moshloop:
@mainawycliffe The spec is saving correctly now, but when trying to edit there is no form and and an empty spec field

mainawycliffe:
@moshloop rebased

PR #1142, opened 2023-05-17
1122-settings-log-backends

### LINKED ISSUE #1122: Settings -> Log Backends

mainawycliffe:
Closes #1122


moshloop:
@mainawycliffe  can we change the icon to the same as the main logs page + change to "Logs" rather than Logging Backend

PR #1161, opened 2023-05-22
Fix-logs-not-showing

mainawycliffe:


mainawycliffe:
@moshloop here is one with logs:

https://deploy-preview-1161--goofy-euclid-75956c.netlify.app/logs?topologyId=018842c0-9a89-d789-b223-2e08d61aff27&logsSelector=Kubernetes

moshloop:
@mainawycliffe looking better - We seem to have lost the time range?

- Also when selecting a component with no log selectors (which shouldn't occur) there is a 400 and the skeleton doesn't hide
- I think the view needs to be updated  as its returning components without log selectors - The list also needs to be alphabetical

mainawycliffe:
@moshloop I have fixed the loading selector issue, and yeah the view needs to be updated or there is an issue that's causing some components not to show selectors.

src/components/FilterLogs/LogsSelector.tsx
```diff
@@ -54,8 +59,8 @@ export default function LogsSelectorDropdown({ classNames }: Props) {
         prefix="Logs selector:"
```
moshloop, line 59:
```suggestion
        prefix="Selector:"
```

src/components/FilterLogs/LogsSelector.tsx
```diff
@@ -54,8 +59,8 @@ export default function LogsSelectorDropdown({ classNames }: Props) {
...
         name="component"
         className="w-auto max-w-[400px] capitalize"
```
moshloop, line 61:
```suggestion
        className="w-auto max-w-[400px]"
```

PR #1162, opened 2023-05-22
1137-show-agent-name

### LINKED ISSUE #1137: Show agent name
- Cards (just under title)
- config, add a column
- health add a tag

mainawycliffe:
Closes #1137

### DIFF
--- a/src/api/services/topology.ts
+++ b/src/api/services/topology.ts
@@ -143,6 +143,13 @@ export const getAgentByID = async (id: string) => {
   return res.data?.[0] ?? null;
 };
 
+export const getAgentByIDs = async (ids: string[]) => {
+  const res = await IncidentCommander.get<AgentItem[] | null>(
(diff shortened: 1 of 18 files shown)


moshloop:
@mainawycliffe - Can we treat the local or '0000' as effectively null or empty ?

mainawycliffe:
@moshloop sure

src/api/services/topology.ts
```diff
@@ -130,6 +130,26 @@ export const getTopologyComponentsWithLogs = () => {
...
 };
 
+type AgentItem = {
+  id: string;
+  name: string;
+  description: string;
+};
+
+export const getAgentByID = async (id: string) => {
+  const res = await IncidentCommander.get<AgentItem[] | null>(
```
moshloop, line 140: Is this going to be cached ?

mainawycliffe: Yeah, we are using react query, so it should be cached and only fetched once, even if you send multiple requests at once.

PR #1175, opened 2023-05-24
1158-config-insights-type-severity-filter-dropdown-items-incorrect

### LINKED ISSUE #1158: Config Insights Type and Severity Filter Dropdown Items are incorrect
image

image

mainawycliffe:
Closes #1158


src/components/ConfigAnalysis/ConfigAnalysisSeverityDropdown.tsx
```diff
@@ -3,18 +3,42 @@ import { Control } from "react-hook-form";
...
+  HiInformationCircle,
+  HiOutlineChevronDoubleDown,
+  HiOutlineChevronDown,
+  HiOutlineMinus
 } from "react-icons/hi";
 import { defaultSelections } from "../Incidents/data";
 import { ReactSelectDropdown } from "../ReactSelectDropdown";
 
 export const configAnalysisSeverityItems = {
+  Low: {
```
moshloop, line 15: Can we not reuse the same constants as the incident severities ?

mainawycliffe: In config insights, we seem to have an extra severity - info, that doesn't exist in incidents severity, other than that, we can.

moshloop:
@mainawycliffe 
- The info, critical, warning icons do not seem to be consistent, 
- The dropdowns should be sorted
- Can we change the font-size of all dropdown to text-sm ?

moshloop:
@mainawycliffe  I don't see the size taking effect - I think react select is adding its own size

PR #1185, opened 2023-05-29
1145-displaying-config-item-crashes-config-not-jsonyaml

### LINKED ISSUE #1145: Displaying config item crashes if the config is not a JSON/YAML.
I tried this with 2 different config_items
- file scraper that scrapes `/etc/hosts` & `/etc/vimrc`
- kubernetes scraper that scrapes `/var/lib/postgresql/data/pgdata/postgresql.conf`

```json
{
    "format": "text",
    "content": "\" All system-wide defaults are set in $VIMRUNTIME/archlinux.vim (usually just\n\" /usr/share/vim/vimfiles/archlinux.vim) and sourced by the call to :runtime\n\" you can find below.  If you wish to change any of those settings, you should\n\" do it in this file (/etc/vimrc), since archlinux.vim will be overwritten\n\" everytime an upgrade of the vim packages is performed.  It is recommended to\n\" make changes after sourcing archlinux.vim since it alters the value of the\n\" 'compatible' option.\n\n\" This line should not be removed as it ensures that various options are\n\" properly set to work with the Vim-related packages.\nruntime! archlinux.vim\n\n\" If you prefer the old-style vim functionalty, add 'runtime! vimrc_example.vim'\n\" Or better yet, read /usr/share/vim/vim80/vimrc_example.vim or the vim manual\n\" and configure vim to your own liking!\n\n\" do not load defaults if ~/.vimrc is missing\n\"let skip_defaults_vim=1\n"
}
```

```
YAMLParseError: Unexpected scalar at node end at line 2, column 3:

" /usr/share/vim/vimfiles/archlinux.vim) and sourced by the call to :runtime
" …
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    at B.onError (composer.js:67:34)
    at r (resolve-end.js:29:21)
    at o (resolve-flow-scalar.js:33:16)
    at A (compose-scalar.js:9:11)
    at I (compose-node.js:22:20)
    at compose-doc.js:30:11
    at B.next (composer.js:145:29)
    at next (<anonymous>)
    at B.compose (composer.js:129:25)
    at compose.next (<anonymous>)
```

mainawycliffe:
Closes #1145


mainawycliffe:
@adityathebe can you confirm this fixes the issue you experience in #1145?

adityathebe:
> @adityathebe can you confirm this fixes the issue you experience in #1145?

Works now 👍

PR #1187, opened 2023-05-29
1134-crd-linked-config-scrapers-not-editable

### LINKED ISSUE #1134: CRD linked config scrapers should not be editable

mainawycliffe:
Closes #1134


moshloop:
@mainawycliffe It display "cantEditMessage" - The delete button should also be removed.

PR #1252, opened 2023-06-17
1236-allow-adding-people-responders

### LINKED ISSUE #1236: Allow adding people as responders
+ assign yourself shorcut

mainawycliffe:
Closes #1236

### DIFF
--- a/src/api/services/responder.ts
+++ b/src/api/services/responder.ts
@@ -1,18 +1,103 @@
 import { IncidentCommander } from "../axios";
 import { resolve } from "../resolve";
+import { Team } from "./teams";
+import { User } from "./users";
+
+/**
+ * column "id" {
+    null    = false
+    type    = uuid
+    default = sql("generate_ulid()")
+  }
+  column "incident_id" {
+    null = false
+    type = uuid
+  }
+  column "type" {
+    null = false
+    type = text
+  }
+  column "index" {
+    null = true
+    type = smallint
+  }
+  column "person_id" {
+    null = true
+    type = uuid
+  }
+  column "team_id" {
+    null = true
+    type = uuid
+  }
+  column "external_id" {
+    null = true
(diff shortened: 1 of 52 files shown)


moshloop:
@mainawycliffe  for the popup can we do:

```
Person: {dropdown}
     OR
Team: {Dropdown} 
```

Rather than a switcher.

Also the new responder isn't immediately added to the panel, nor do I seem them on the incidents list page

mainawycliffe:
@moshloop like this:

image

moshloop:
@mainawycliffe  - Without the inline Label and with an `OR` text between them 
The modal title also needs to be changed

moshloop:
@mainawycliffe    can we do this vertically rather than horizontal with the labels on the left instead of the top ?

mainawycliffe:
@moshloop sure, in terms of the responder column, I am working on displaying them, might need to do some refactoring around.

moshloop:
@mainawycliffe - this is better

- Can we move the add responders to after the list of responders
- The avatar icon doesn't seem consistent between incident changelog (what it should be) and the responders list in both details and the incident list

src/api/services/responder.ts
```diff
@@ -1,18 +1,103 @@
 import { IncidentCommander } from "../axios";
 import { resolve } from "../resolve";
+import { Team } from "./teams";
+import { User } from "./users";
+
+/**
```
moshloop, line 6: remove this comment

moshloop:
I think we should also be using the incident_summary view: https://github.com/flanksource/duty/blob/main/views/009_incident_views.sql#L56

moshloop:
@mainawycliffe Can we make the avatars rounded and sm by default so they are consistent across the board?

mainawycliffe:
@moshloop like this:

image

PR #1254, opened 2023-06-19
feat: added new connection forms for Shoutrrr

adityathebe:
* added support for Formik hints
* Also resolves: https://github.com/flanksource/flanksource-ui/issues/1256


moshloop:
@adityathebe  can we we add the query params as well ?

adityathebe:
> @adityathebe can we we add the query params as well ?

@moshloop The query params are notifications only. I don't think they should be stored with the connection.

src/icons/make.sh
```diff
@@ -1,18 +1,50 @@
...
+# Parameters:
+#   $1: The string to check
+# Returns:
+#   0 (success) if the string matches a JavaScript keyword
+#   1 (failure) otherwise
+function is_javascript_keyword() {
+  local keyword=" $1 " # Add spaces around the keyword for exact matching
+
+  # List of JavaScript keywords
+  local javascript_keywords=(
```
adityathebe, line 14: one of the icons is named "switch", so I had to add this check

Issue #1238, opened 2023-06-13
Config insights panel should be sored by severity, then alphabetically


moshloop:

yashmehrotra:
Fixed via: https://github.com/flanksource/duty/pull/188

PR #1255, opened 2023-06-20
chore: add sorting order to configs side panel

yashmehrotra:
Fixes: https://github.com/flanksource/flanksource-ui/issues/1238


mainawycliffe (changes requested):
@yashmehrotra Sorry, will be away for most of the week, but had a quick glance, looks okay, but due to perfomance reason, the best place would be in the useQuery, add a select property in the 3rd argument and return the sorted data or useMemo.

mainawycliffe (approved):
LGTM

PR #1276, opened 2023-06-29
fix: fix topology insights not loading

### LINKED ISSUE #1242: Topology Insights is Broken
The `InsightsDetails` details component currently only supports configs, instead of both topology and configs, meaning it's broken for the UI.

image

mainawycliffe:
Closes #1242

### DIFF
--- a/src/api/services/configs.ts
+++ b/src/api/services/configs.ts
@@ -323,7 +323,7 @@ export const getConfigsChangesTypesFilter = async () => {
   return res.data;
 };
 
-export const getConfigInsights = async <T>(
+export const getConfigInsights = (
   configId: string,
   pageIndex?: number,
   pageSize?: number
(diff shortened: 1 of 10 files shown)


src/api/services/configs.ts
```diff
@@ -335,7 +335,7 @@ export const getConfigInsights = async <T>(
     }`;
   }
   return resolve(
-    ConfigDB.get<T>(
+    ConfigDB.get<ConfigTypeInsights[]>(
       `/config_analysis?select=*,config:configs(id,name,config_class,type)&config_id=eq.${configId}${paginationQueryParams}`,
```
moshloop, line 339: We shouldn't be getting all columnes

src/api/services/configs.ts
```diff
@@ -335,8 +335,8 @@ export const getConfigInsights = async <T>(
     }`;
   }
   return resolve(
-    ConfigDB.get<T>(
-      `/config_analysis?select=*,config:configs(id,name,config_class,type)&config_id=eq.${configId}${paginationQueryParams}`,
+    ConfigDB.get<ConfigTypeInsights[]>(
+      `/config_analysis?select=id,source,analyzer,analysis_type,message,severity,status,analysis,first_observed,config:configs(id,name,config_class,type)&config_id=eq.${configId}${paginationQueryParams}`,
```
moshloop, line 339: We shouldn't be including analysis as its pretty large - It should only show on the popup window

PR #1282, opened 2023-07-03
flatten the topology search when filtering by team

adityathebe:


src/pages/TopologyPage.tsx
```diff
@@ -104,13 +104,17 @@ export function TopologyPage() {
...
+        flatten: false,
         hidden: showHiddenComponents === "no" ? false : undefined
       };
+
+      if (
+        topologyType.toLowerCase() !== "all" ||
+        team.toLowerCase() !== "all"
+      ) {
+        apiParams.flatten = true;
+      }
```
mainawycliffe, line 116:
```suggestion
        hidden: showHiddenComponents === "no" ? false : undefined,
        ...((topologyType.toLowerCase() !== "all" ||
          team.toLowerCase() !== "all") && {
          flatten: true
        })
```

mainawycliffe: I would probably do it like this but otherwise looks good.

mainawycliffe (approved):
Otherwise looks good.

adityathebe:
We won't be flattening filters by team

PR #1285, opened 2023-07-03
1241-unable-edit-incident-titles

### LINKED ISSUE #1241: Unable to edit incident titles

mainawycliffe:
Closes #1241


moshloop:
@mainawycliffe - can we do this as inline edit like the hypothesis titles

PR #1287, opened 2023-07-04
feat: add redis and mongo forms

### LINKED ISSUE #880: Form - Health - Mongo

### LINKED ISSUE #881: Form - Health - Redis

yashmehrotra:
Fixes #880 and Fixes #881

### DIFF
--- a/src/components/Forms/Health/MongoHealthFormEditor.tsx
+++ b/src/components/Forms/Health/MongoHealthFormEditor.tsx
@@ -1,0 +1,86 @@
+import FormikTextInput from "../Formik/FormikTextInput";
+import FormikTemplateFields from "../Formik/FormikTemplateFields";
+import FormikAuthFieldsGroup from "../Formik/FormikAuthFieldsGroup";
+import FormikCheckboxFieldsGroup from "../Formik/FormikCheckboxFieldsGroup";
+import FormikIconPicker from "../Formik/FormikIconPicker";
+import { getIn, useFormikContext } from "formik";
+import { useEffect } from "react";
+import FormikScheduleField from "../Formik/FormikScheduleField";
+import FormikEnvVarConfigsFields from "../Formik/FormikConfigEnvVarFields";
+import FormikConnectionField from "../Formik/FormikConnectionField";
+
+type MongoHealthFormEditorProps = {
+  fieldName: string;
+  specsMapField: string;
+};
+
+export function MongoHealthFormEditor({
+  fieldName: name,
+  specsMapField
+}: MongoHealthFormEditorProps) {
+  const { values, setFieldValue } = useFormikContext();
+
+  const fieldName = `${name}.${specsMapField}`;
+
+  const nameValue = getIn(values, `${fieldName}.name`);
+
+  // when name changes, we want to update the name of the top level field
+  useEffect(() => {
+    setFieldValue("name", nameValue);
+  }, [nameValue, setFieldValue]);
+
+  return (
+    <>
+      <div className="flex flex-row md:space-x-2">
+        <FormikTextInput
+          className="flex flex-col w-1/2"
+          name={`${fieldName}.name`}
+          label="Name"
+          required
+        />
+        <FormikIconPicker
+          className="flex flex-col w-1/2"
+          name={`${fieldName}.icon`}
+          label="Icon"
+        />
+      </div>
+
+      <FormikScheduleField name={`${name}.schedule`} />
+
+      <FormikCheckboxFieldsGroup name={`${name}.username`} label="Username">
+        <FormikEnvVarConfigsFields name={`${fieldName}.username`} />
+      </FormikCheckboxFieldsGroup>
+
+      <FormikCheckboxFieldsGroup name={`${name}.password`} label="Password">
+        <FormikEnvVarConfigsFields name={`${fieldName}.password`} />
+      </FormikCheckboxFieldsGroup>
+
+      <FormikConnectionField
+        name={`${fieldName}.connection`}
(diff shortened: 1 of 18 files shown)


src/components/Forms/Health/RedisHealthFormEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+      <FormikTextInput name={`${fieldName}.description`} label="Description" />
+
+      <FormikTextInput
+        name={`${fieldName}.addr`}
+        label="Address"
+        required
+        hint="host:port"
+      />
+      <FormikTextInput
```
mainawycliffe, line 56: Quick question: Is this supposed to be a number field only? 

image

src/components/Forms/Health/RedisHealthFormEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+          name={`${fieldName}.icon`}
+          label="Icon"
+        />
+      </div>
+
+      <FormikScheduleField name={`${name}.schedule`} />
+
+      <FormikTextInput name={`${fieldName}.description`} label="Description" />
+
+      <FormikTextInput
```
mainawycliffe, line 50: May be make the hint more descriptive or give an example.

image

mainawycliffe (commented):
Just a couple of nitpicking comments otherwise looks good.

src/components/Forms/Health/RedisHealthFormEditor.tsx
yashmehrotra: Yeah, redis databases are numbers

src/components/Forms/Health/RedisHealthFormEditor.tsx
yashmehrotra: Can you suggest a better description ?

src/components/Forms/Health/MongoHealthFormEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+          className="flex flex-col w-1/2"
+          name={`${fieldName}.icon`}
+          label="Icon"
+        />
+      </div>
+
+      <FormikScheduleField name={`${name}.schedule`} />
+
+      <FormikTextInput name={`${fieldName}.username`} label="Username" />
+      <FormikTextInput name={`${fieldName}.password`} label="Password" />
```
moshloop, line 49: These should be env var fields + we need a connection selection field

moshloop: Also missing the url field

mainawycliffe: @moshloop for the connection field, is it a dropdown connection to the connection view?

moshloop:
I think the required field on the db should be removed, if not specified its a 0

src/components/Forms/Health/RedisHealthFormEditor.tsx
```diff
@@ -0,0 +1,86 @@
...
+        required
+        hint="Address to redis server in the format of host:port"
+      />
+
+      <FormikTextInput
+        name={`${fieldName}.db`}
+        label="Database"
+        type="number"
+        hint="Database to be selected after connecting to the server"
+        required
```
moshloop, line 62:
```suggestion
```

moshloop:
@mainawycliffe on the Redis side the database should default to `0` and not be required
On the mongo side we have duplicated user/pass

We need to update the Auth component so that NTLM and NTLM2 optional e.g. only specified on the http check as:

```jsx
<Auth 
   types={{[
  {label: "NTLM",  value: {ntlm: true}},
  {label: "NTLMv2",  value: {ntlm2: true}},
]}}
```

We might then reuse it like:


```jsx
<Auth 
   types={{[
  {label: "Scram",  value: {authType: "scram"}},
  {label: "Digest",  value: {authType: "Digest"}},
]}}
```

For the EnvVarValue selector we also need options to `hideName` and `hideCheckbox` which would then render as:

```

(When Value is selected)
Username:   [Value] [Secret] [Configmap] 
                    <input> 
(When config/secret selected)
Username:   [Value] [Secret] [Configmap] 
                 Name: <input> Key: <input>       
```

mainawycliffe:
> For the EnvVarValue selector we also need options to `hideName` and `hideCheckbox` which would then render as:
> 
> ```
> 
> (When Value is selected)
> Username:   [Value] [Secret] [Configmap] 
>                     <input> 
> (When config/secret selected)
> Username:   [Value] [Secret] [Configmap] 
>                  Name: <input> Key: <input>     
> ```

@moshloop I don't understand what you mean here?

mainawycliffe:

moshloop:
@mainawycliffe - I think lets rather do:
Username:   
 <**Static**>|<Config Map>|<Secret>.  [ _______________ ]
Password:
Name  [ _______________ ] Key  [ _______________ ]  <Static>|<Config Map>|<**Secret**>

- These changes should also apply to all other auth fields 
- NTLM and NTLMv2 are only application to HTTP and should be hidden for others with something like:

```
<Auth 
   types={{[
  {label: "NTLM",  value: {ntlm: true}},
  {label: "NTLMv2",  value: {ntlm2: true}},
]}}
```

moshloop:
@mainawycliffe why do we have checkboxes now? It should be Label, radiobuttongroup, fields ?

moshloop:
@mainawycliffe we still seem to have a scrolling issue on the modal header ?

mainawycliffe:
@moshloop the modal scrolls on the health page should be fixed now

PR #1311, opened 2023-07-10
1294-incident-rules-form

### LINKED ISSUE #1294: Incident rules form

mainawycliffe:
Closes #1294


moshloop:
@mainawycliffe 

- the components are a list - and at least 1 is required
- component type should be dropdown of component types
- Selector should just be an input
- Only title, description, type (dropdown), severity (dropdown) - with only description being optional
-

PR #1333, opened 2023-07-25
feat: add page and form for notifications

### LINKED ISSUE #1293: Notifications form

mainawycliffe:
### TODO

- [x] Add Person/Team/Custom Connection Column
- [x] Fix `deleted_at` not working
- [ ] Clarify on custom services format

Closes #1293

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,7 +69,7 @@
         "react-loading-icons": "^1.0.8",
         "react-mentions": "^4.4.0",
         "react-router-dom": "^6.2.1",
-        "react-select": "^5.2.2",
+        "react-select": "^5.7.4",
         "react-table": "^7.7.0",
         "react-toastify": "^9.0.8",
         "react-tooltip": "^4.2.21",
@@ -3414,7 +3414,6 @@
       "version": "1.4.1",
       "resolved": "https://registry.npmjs.org/@floating-ui/core/-/core-1.4.1.tgz",
       "integrity": "sha512-jk3WqquEJRlcyu7997NtR5PibI+y5bi+LS3hPmguVClypenMsCY3CBa3LAQnozRCtCrYWSEtAdiskpamuJRFOQ==",
-      "dev": true,
       "dependencies": {
         "@floating-ui/utils": "^0.1.1"
       }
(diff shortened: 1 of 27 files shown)


adityathebe:
image

@moshloop @mainawycliffe 

- I think notification id column can be removed.
- There can be multiple recipients (one person at max and one team at max and as many custom ones). I've chosen both team and person for some notifications but only person is showing up.
- templates can probably be removed as well since they can be very long?

adityathebe:
image

- URL or connection should probably an exclusive field. Mandatory but only one can be provided.
- template field could be a text area field instead.

mainawycliffe:
> image
> 
> @moshloop @mainawycliffe
> 
> * I think notification id column can be removed.
> * There can be multiple recipients (one person at max and one team at max and as many custom ones). I've chosen both team and person for some notifications but only person is showing up.
> * templates can probably be removed as well since they can be very long?

@adityathebe is custom ones an array then?

adityathebe:
@mainawycliffe ya it's an array

mainawycliffe:
@moshloop for persons list, the add responder modal is small, so we can get way without needing to handle overflow, hence why the persons list and teams list don't cause scrolls, which isn't the case with the notifications forms.

On the first issue, it stems from missing notification id from the notifications summary view, which I have created in a PR (https://github.com/flanksource/duty/pull/261)

moshloop:
@mainawycliffe as a rule dropdown should always be above a modal body and never inside.

moshloop:
@mainawycliffe - I still cannot edit, and when saving I get a malformed array error

mainawycliffe:
@moshloop there are still some changes yet to be deployed on the backend, following up on that, as the id is need to edit and haven't been able to test it yet.

moshloop:
@mainawycliffe  

- [ ] After deleting a notification the modal doesn't close
- [ ] Remove the template  field from table
- [ ] Show the filter in the table
- [ ] Add a subject field
- [ ] The status field should use the same component as the other tables (in future we will add click to open more details)
- [ ] Properties should not be duplicated under other
- [ ] The template text area should be 5 lines of height
- [ ] The template text area type should be go template html without any schema

mainawycliffe:
@moshloop I have added title to the form and table, but not showing as title is missing in the notification summary, which I have added using the following [PR](https://github.com/flanksource/duty/pull/269).

adityathebe:
@mainawycliffe I think the events on the dropdown should be identical to the ones in the table.

image

mainawycliffe:
@adityathebe fixed the issue

adityathebe:
image

Title is also an optional field since we have defaults.

mainawycliffe:
@adityathebe let me investigate the issue and fix the other one and PR.

adityathebe:
npm install fixed the freezing issue. Also just noticed, we are missing events for components. They were recently added

```
EventComponentStatusHealthy   = "component.status.healthy"
EventComponentStatusUnhealthy = "component.status.unhealthy"
EventComponentStatusInfo      = "component.status.info"
EventComponentStatusWarning   = "component.status.warning"
EventComponentStatusError     = "component.status.error"
```

PR #1373, opened 2023-09-09
Disabled Incidents and Logs Features using Feature Flags

### LINKED ISSUE #1360: Disable incidents feature flag
- Hide Incidents, rules tabs
- Hide all Attach to incidents buttons

### LINKED ISSUE #1366: Disable logs feature flag
- Remove logs and log backends navigation
- Remove View logs action

mainawycliffe:
Closes #1360
Closes #1366

### DIFF
--- a/src/components/Canary/minimal.tsx
+++ b/src/components/Canary/minimal.tsx
@@ -1,20 +1,18 @@
-import React, { useCallback, useEffect, useMemo, useState } from "react";
-import { Modal } from "../Modal";
-import { CheckDetails } from "./CanaryPopup/CheckDetails";
-import { CheckTitle } from "./CanaryPopup/CheckTitle";
-import { CanaryCards } from "./card";
-import { CanaryTable } from "./table";
+import React, { useCallback, useEffect, useState } from "react";
 import { useSearchParams } from "react-router-dom";
 import { EvidenceType } from "../../api/services/evidence";
-import { AttachEvidenceDialog } from "../AttachEvidenceDialog";
-import { isCanaryUI } from "../../context/Environment";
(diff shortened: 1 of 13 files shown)


src/components/Canary/minimal.tsx
```diff
@@ -142,7 +153,7 @@ const MinimalCanaryFC = ({
             {selectedCheck?.canary_id && (
               <HealthCheckEdit check={selectedCheck as HealthCheck} />
             )}
-            {!isCanaryUI && (
+            {!isCanaryUI && !isFeatureDisabled && (
```
moshloop, line 156: Can we encapsulate this in a AttachAsEvidenceButton that uses the features context inside (with many different styles for primary, toolbar, menu) ?

src/components/Canary/minimal.tsx
```diff
@@ -107,23 +116,25 @@ const MinimalCanaryFC = ({
...
-          includeMessages: true,
-          start: timeRange
-        }}
-        type={EvidenceType.Check}
-        callback={(success: boolean) => {
-          console.log(success);
-        }}
-      />
+      {!isIncidentManagementFeatureDisabled && (
+        <AttachEvidenceDialog
```
moshloop, line 120: Can we move the feature check inside the dialog ?

src/components/Canary/minimal.tsx
mainawycliffe: I have created the button, only one place where moving to the button is trick -> the Topology Dropdown, as the dialog needs to exist outside the dropdown where the button is, otherwise, it doesn't load, since the dialog gets removed from the DOM when the dropdown gets closed.

PR #1381, opened 2023-09-12
1334-config-list-filtering-ignores-type-selection

### LINKED ISSUE #1334: Config list filtering ignores type selection

mainawycliffe:
Closes #1334


src/context/TopologyPageContext.tsx
```diff
@@ -1,5 +1,3 @@
-import React, { useState, createContext, useContext } from "react";
```
moshloop, line 1: Why are we removing this ?

mainawycliffe: It's no longer being used anywhere, we currently use React Query for caching purposes.

PR #1382, opened 2023-09-13
fix: incorrect mapping for Slack connection & notification template form

adityathebe:
- Slack connection form was incorrectly saving the "BotName" field as "fromName"
- the notification form should only support markdown templates.


.tool-versions
```diff
@@ -1 +1 @@
-nodejs 16.14.1
+nodejs 18.17.1
```
mainawycliffe, line 1: Let's remove this from the commits, may be add it to gitignore

mainawycliffe (commented):
Just a single comment on the `.tool-versions` otherwise looks good

PR #1395, opened 2023-09-14
fix: don't return relative date if it's 0001-01-01T00:00:00+00:00

### LINKED ISSUE #1394: CRD config scrapers have created date 2000y ago

mainawycliffe:
Closes #1394


mainawycliffe:
@moshloop the reason for this issue is that the dates are 0001-01-01T00:00:00+00:00 from the DB, so for now I have just masked them, but probably an issue when creating that needs investigating.

PR #1444, opened 2023-10-17
fix: decode logs in results object and remove overflow of caused by text

### LINKED ISSUE #1436: Playbook Actions Results Improvements
- [ ] For Results with logs, decode the logs before display (ANSI)
- [ ] Do not overflow the results section and instead wrap the content

mainawycliffe:
Closes #1436


moshloop:
@mainawycliffe  we are missing the handling of errors: https://deploy-preview-1444--flanksource-demo-stable.netlify.app/playbooks/runs/018af987-d406-a5df-5a63-a07d8842c7cf

Also, I don't think we should be downloading the result/error on load - only when the step is selected

PR #1459, opened 2023-10-24
feat: add new git connection type for connections

### LINKED ISSUE #1449: Add a new "git" connection type
Ref: https://github.com/flanksource/duty/pull/303

mainawycliffe:
Closes #1449

### DIFF
--- a/src/components/Connections/connectionTypes.tsx
+++ b/src/components/Connections/connectionTypes.tsx
@@ -91,18 +91,29 @@ export type ConnectionType = {
   forNotification?: boolean;
 };
 
+export const commonConnectionFormFields: ConnectionFormFields[] = [
+  {
+    label: "Name",
+    key: "name",
+    type: ConnectionsFieldTypes.input,
+    required: true
+  },
+  {
+    label: "Namespace",
+    key: "namespace",
+    type: ConnectionsFieldTypes.input,
+    default: "default",
+    required: true
+  }
+];
+
 export const connectionTypes: ConnectionType[] = [
   {
     title: "Postgres",
     icon: "postgres",
     value: ConnectionValueType.Postgres,
     fields: [
-      {
-        label: "Name",
-        key: "name",
-        type: ConnectionsFieldTypes.input,
-        required: true
-      },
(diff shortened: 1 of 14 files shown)


src/components/Connections/connectionTypes.tsx
```diff
@@ -1807,6 +1807,55 @@ export const connectionTypes: ConnectionType[] = [
...
+    value: ConnectionValueType.Git,
+    fields: [
+      {
+        label: "Name",
+        key: "name",
+        type: fieldTypes.input,
+        required: true
+      },
+      {
+        label: "Certificate",
```
moshloop, line 1823: We need a username and password as well and rename the certificate label to SSH Key
+ reorder name, url, user, pass, certificate, ref

src/components/Connections/connectionTypes.tsx
```diff
@@ -1807,6 +1807,69 @@ export const connectionTypes: ConnectionType[] = [
...
+        required: true
+      },
+      {
+        label: "URL",
+        key: "url",
+        type: fieldTypes.input,
+        required: true
+      },
+      {
+        label: "Username",
```
moshloop, line 1829: Its either Username or Certificate, not both and I don't think any of them are required.

moshloop:
@mainawycliffe I can't see to save the form ?

mainawycliffe:
@moshloop  Let me investigate

moshloop:
@mainawycliffe - The password doesn't seem to save - also if you specify a user/pass and then save and  edit it shows you an empty ssh key, it should choose the value that is not null 
- i  should also default to user/pass not ssh

moshloop:
Can we also update the user/pass style to the same as on the http form (extracting and resuing components as needed)

moshloop:
@mainawycliffe  We now have a Username None/Basic that expands to username and password under username?  It should just be Auth: None | Basic | SSH,

moshloop:
@mainawycliffe  We also need to add a namespace (defaulting to `default`) to all the connection forms and the table

PR #1461, opened 2023-10-24
Refactoring

moshloop:


mainawycliffe (approved):
For Types, I think we ought to automate that at some points via Swagger to Typescript type or Postgress to Typescript, I will file an issue for this.

PR #1465, opened 2023-10-26
1451-connection-properties-need-stored-mapstring-string

### LINKED ISSUE #1451: Connection properties need to be stored as map<string, string>
Eg: For aws connection,

```json
{
  "region": "",
  "profile": "",
  "insecureTLS": false
}
```

`insecureTLS` should be "false".

We treat connection properties as map[string]string in Go and this caused an error when unmarshalling.

mainawycliffe:
Closes #1451


moshloop:
@mainawycliffe Is this not something we can in global presubmit handling ?

mainawycliffe:
May be with a hook before sending data to the backend, this is very specific to properties and the data as currently structured in the form, doesn't have a property field and get converted by the presubmit hook where necessary.

PR #1470, opened 2023-11-01
1453-treat-check-names-grouping

### LINKED ISSUE #1434: Canary table - grouping latency / uptime
When rows are grouped in uptime, do x/y as in 6/6 or 100%, and for latency use the 95% of all of rows

### LINKED ISSUE #1453: Treat '/' in check names as grouping
.e.g. instead of 

```
PodOOMKilled/namespace/podid1
PodOOMKilled/namespace/podid2
```

do

```
PodOOMKilled
      namespace
          podid1
          podid2
`

mainawycliffe:
Closes #1434
Closes #1453

### DIFF
--- a/src/components/Canary/CanaryTableColumns.tsx
+++ b/src/components/Canary/CanaryTableColumns.tsx
@@ -1,0 +1,134 @@
+import { CellContext, ColumnDef } from "@tanstack/react-table";
+import { HealthCheck } from "../../api/types/health";
+import { IoChevronForwardOutline } from "react-icons/io5";
+import {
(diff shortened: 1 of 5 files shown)


moshloop:
@mainawycliffe  We need to handle the case of a single item - it strips off the prefix, but then then there is no group header

moshloop:
@mainawycliffe  We need to handle the case of a single item - it strips off the prefix, but then then there is no group header

there is also groups with `2 /  0`  - we cal also remove the spacing so thats its just `2/2`

PR #1488, opened 2023-11-08
1477-gauge-property-renderer

### LINKED ISSUE #1477: Gauge  property renderer
Where the max value is known render a 
Screenshot 2023-11-03 at 09 36 04

On hover on the progress bar show the tooltip "57% of <total>

+ 

In the details panel re-render the headline properties as they appear on the cards

mainawycliffe:
Closes #1477

### DIFF
--- a/src/components/Agents/AgentName.tsx
+++ b/src/components/Agents/AgentName.tsx
@@ -19,5 +19,12 @@ export default function AgentName({ agentId }: TopologyCardAgentProps) {
     return null;
   }
 
-  return <Badge title={agent.description} text={agent.name} />;
+  return (
+    <Badge
+      colorClass="bg-gray-100 text-gray-800"
+      size="xs"
+      title={agent.description}
+      text={agent.name}
+    />
+  );
 }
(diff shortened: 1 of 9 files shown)


moshloop:
1) It seems there is an issue with < 2% rendering
Screenshot 2023-11-10 at 15 25 13
2) Reduce the height to 1.5
3. The gauge is in addition to the current value, not instead of
4. Seem to have lost unit formatting on those pods without a max value
5. The bar should never go above 100% (See https://deploy-preview-1488--goofy-euclid-75956c.netlify.app/topology/0189f817-9d20-ff3f-718f-969dc41c0ce0?sortBy=status&sortOrder=desc&refererId=018b81ad-0206-8473-1282-014a3dd197c5) although in that case I don't see why its calculated a percentage of 5500%

moshloop:
@mainawycliffe ^

moshloop:
@mainawycliffe Can we left align the value and the bar in the side panel? I think also the font color is not consistent with other headlines

PR #1501, opened 2023-11-14
fix: improve responsiveness of the canary UI

mainawycliffe:


mainawycliffe:

PR #1514, opened 2023-11-22
1510-topology-property-improvements

### LINKED ISSUE #1510: Topology property improvements
- [x] Render the headline at the top of the panel, (and skip the headlines below)
- [x] Don't titlecase property names
- [x] Allow property grouping by either `:` or `/`  (and split on the last index of)
- [x] Move deleted at to created/updated line (and remove the `at` on all 3
- [x] Order properties alphabetically (first by group, then by name in group)
- [x] If there is a namespace label, prefix the name with it e.g. `namespace / name`
- [x] When there is no cost, don't render the label
- [x] Apply this style to grouped proeprties `<div class="flex flex-col gap-2 mx-1 px-2" style="
    border-left: gray 1px dotted;`
- [x] Make type line be `AWS::EKS::Cluster created by mission-control/aws` - the scraper link having normal url color / underline 
- [x] Apply heuristics for rendering some properties into 2 columns - note this applies to properties after the group prefix has been removed
       * region / zone
       * vpc / subnet
       * os / arch
       * version / part-of

mainawycliffe:
Closes #1510

### DIFF
--- a/src/api/types/configs.ts
+++ b/src/api/types/configs.ts
@@ -33,10 +33,10 @@ export interface Costs {
 
 export function isCostsEmpty(costs: Costs) {
   return (
-    costs.cost_per_minute === undefined &&
-    costs.cost_total_1d === undefined &&
-    costs.cost_total_7d === undefined &&
-    costs.cost_total_30d === undefined
+    !costs.cost_per_minute &&
+    !costs.cost_total_1d &&
+    !costs.cost_total_7d &&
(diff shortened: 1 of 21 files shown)


moshloop:
- Can we use the `created by` rather than `by` 
- Missing space on catalog page between by and ling
- Property sorting should be case insensitive 
- Move the dashed border on sub properties from top to left
- Flip the label / value styles - label should be lighter, and value should be bolder

PR #1530, opened 2023-11-29
create a dedicated number input, with assertion and validation build in

### LINKED ISSUE #1521: Add validation for thresholdMillis in health check form before sending
The thresholdMillis in health check can only be integer else the canary fails, there are cases when its being set as an empty string in the spec which results in the following error

```
v1.CanarySpec.HTTP: []v1.HTTPCheck: v1.HTTPCheck.ThresholdMillis: readUint64: unexpected character: �, error found in #10 byte of ...|Millis": ""}], "sche|..., bigger context ...|com", "responseCodes": [200], "thresholdMillis": ""}], "schedule": "@every 1m"}|...
```

image

mainawycliffe:
Closes #1521


moshloop:
@mainawycliffe can. you look at the failing tests ?

PR #1531, opened 2023-11-30
feat: add kube options to the add agent

### LINKED ISSUE #1517: Add agent - kubernetes options
Add an option for setting kubernetes scrapper values on the chart:

values:
    kubernetes:
          scraper:
             enabled=true
             interval:
             exclusions:
                      -
                      -

mainawycliffe:
Closes #1517


moshloop:
@mainawycliffe  Enabled should be a selection - it is always true if kubernetes is selected, the interval should also default to 30m and be labelled Scrape Interval, clusterName should be the agent name

moshloop:
@mainawycliffe I don't see the changes we discussed ?

PR #1533, opened 2023-11-30
1524-connection-webhook-cannot-create-to-required-field

### LINKED ISSUE #1511: Connection form minor fixes
- [ ] For SFTP, the port is unspecified a default value should be saved on the URL. Right now `undefined` is being saved instead.
- [ ] I think the password field should not remember previous inputs.
image

### LINKED ISSUE #1523: Connection SMTP - Cannot create due to required field error
- Encryption method / smtp auth method should be checkboxgroup
- Host and port should be on 1 line, just above user/pass
- User/pass is not required
- User should be envvaluefrom field
- 
Screenshot 2023-11-26 at 09 54 57

### LINKED ISSUE #1524: Connection webhook - Cannot create to required field
Request method should also be a checkbogroup

mainawycliffe:
Closes #1524
Closes #1523
Closes #1511


moshloop:
@mainawycliffe  there is an issue with the port - If you try and save it does nothing (and on 1 occasion I saw port is required, even though it had a value) - If you change the value from the default - then it works

moshloop:
@mainawycliffe - I still get an error when saving SFTP

For Webhooks, If I change get/post then the message key says its required ?

mainawycliffe:
@moshloop fixed the issue, it had to do with mapping the db data to form when editing, with message key not mapped correctly.

PR #1535, opened 2023-11-30
feat: add smb connection type

### LINKED ISSUE #1512: I think we're missing SMB connection form as well?

mainawycliffe:
Closes #1512


moshloop:
@mainawycliffe it seems only the workstation gets saved ?

There should only be a server, port (optional) (on the same line) and share,   no folder or workstation - also the required fields should always come first followed by the optionals

mainawycliffe:
@moshloop I have made the changes, but is built on top of the following [PR](https://github.com/flanksource/flanksource-ui/pull/1533), which we should merge first

mainawycliffe:
@moshloop this should be ready to go as well

PR #1537, opened 2023-12-01
feat: add aws s3 connection form

mainawycliffe:


moshloop:
insecure_tls shouldn't be a property its column in connections table, also bucket should be required 
Can we also make the update/save buttons have loading indicator?

moshloop:
@mainawycliffe It doesn't look like insecure_tls  == true reflects correct in the checkbox on edit

PR #1539, opened 2023-12-05
fix: hide intercom chat launcher by default

mainawycliffe:
Still remaining:

- [ ] Show chat launcher is user has messages


moshloop:
@mainawycliffe Can we rename to just "Docs" and to Chat (with a chat icon) - Also this should show before instance provisioning - if there is a failure and that is all the users see then they should be able to reach support from there

moshloop:
@mainawycliffe we are missing the docs icon now ?

PR #1600, opened 2024-01-08
1590-Playbook-Param-Types

### LINKED ISSUE #1590: Playbook Param Types
- YAML|JSON|code - Show a VsCode editor + increase the size of dialog
- Checkbox, Textearea, Combo (with static values) 
- Config / Component / Team / Person selector

mainawycliffe:
Closes #1590


src/api/services/configs.ts
```diff
@@ -123,6 +123,17 @@ export const getConfigChanges = (
   );
 };
 
+export const getConfigList = (type?: string) => {
+  return resolve<Pick<ConfigItem, "id" | "name" | "config_class" | "type">[]>(
```
moshloop, line 127: I think we need icon here as well

mainawycliffe: @moshloop  Don't we use the config_class and type for icons for configs?

PR #1603, opened 2024-01-09
fix: deployment should listen on 0.0.0.0

moshloop:


scottrigby (commented):
Patch works as expected locally 👍 

either port-forward to deployment on port 3000, or to service on port 80 both work.

```
kubectl port-forward deployments/incident-manager-ui :3000
or
kubectl port-forward svc/incident-manager-ui :80
```

PR #1640, opened 2024-01-22
fix: use browser url instead of backend url for agent host name

### LINKED ISSUE #1639: Agent Install - Host name should be browser url no _BACKEND_URL

mainawycliffe:
Fixes #1639


moshloop:
@mainawycliffe please check failing test cases

PR #1657, opened 2024-01-26
1591-Sorting-by-component-property-does-not-take-effect-immiedatly

### LINKED ISSUE #1591: Sorting by component property does not take effect immiedatly
i.e - After choosing a new property, you need to change the sort order for it to actually take effect

mainawycliffe:
Fixes #1591

### DIFF
--- a/src/components/Topology/TopologyPopover/topologySort.tsx
+++ b/src/components/Topology/TopologyPopover/topologySort.tsx
@@ -35,7 +35,7 @@ export function getSortLabels(topology: Topology[]) {
         labels[h.name] = true;
         currentSortLabels.push({
           id: defaultSortLabels.length + index,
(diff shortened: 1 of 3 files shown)


src/components/Topology/TopologyPopover/topologySort.tsx
```diff
@@ -35,7 +35,7 @@ export function getSortLabels(topology: Topology[]) {
         labels[h.name] = true;
         currentSortLabels.push({
           id: defaultSortLabels.length + index,
-          value: (h.name ?? "").toLowerCase(),
+          value: `field:${(h.name ?? "").toLowerCase()}`,
```
yashmehrotra, line 38: I don't think this should be lower case, if someone writes "Memory" etc. in their list this will break, also we should allow use to define custom sorting fields later on properties

PR #1658, opened 2024-01-26
1576-Display-usernamepassword-correctly-when-creating-an-agent

### LINKED ISSUE #1576: Display username/password correctly when creating an agent
Username should TOKEN (which can be hardcoded in chart)

mainawycliffe:
Fixes #1576

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,6 +47,7 @@
         "eslint-config-next": "^14.0.3",
         "eslint-plugin-jest-dom": "^4.0.3",
         "formik": "^2.2.9",
+        "handlebars": "^4.7.8",
         "jest": "^29.3.1",
         "jest-environment-jsdom": "^29.3.1",
         "jotai": "^2.0.3",
@@ -17438,7 +17439,6 @@
       "version": "4.7.8",
       "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.7.8.tgz",
       "integrity": "sha512-vafaFqs8MZkRrSX7sFVUdo3ap/eNiLnb4IakshzvP56X5Nr1iGKAIqdX6tMlm6HcNRIkr6AxO5jFEoJzzpT8aQ==",
-      "dev": true,
       "dependencies": {
         "minimist": "^1.2.5",
         "neo-async": "^2.6.2",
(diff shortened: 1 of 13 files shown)


adityathebe:
image

@mainawycliffe The helm install command is just duplicated here. I think when the `kubernetes` checkbox is unchecked we do not need the second helm install command.

image

Likewise, the second helm release needs to be removed from the flux manifests.
---
And maybe we need to add a padding below the warning message.

mainawycliffe:
@adityathebe got you, will make the change.

adityathebe:
@mainawycliffe also just noticed, the kubernetes exclusions are not propagated to the helm release here
image

I do see them in here though

image

and I don't think we need the first `helm install` here

moshloop:
@mainawycliffe  None of the durations should have `in` - For the sleeping actions / tasks then it should be a `-` with maybe a tooltip over the the `-` or calendar icon

adityathebe:
@mainawycliffe I'm still seeing the duplicate helm install & helm release for mission-control-kubernetes even when the kubernetes checkbox is unchecked.

adityathebe:
This `exclusions` should be part of the values in the `mission-control-kubernetes` helm release.

image


`spec.values.scraper.exclusions.names`

@moshloop we have [multiple exclusions for kubernetes](https://github.com/flanksource/mission-control-registry/blob/c1bcf2f39b177f5e9972507dcda88c77c96c3af4/charts/kubernetes/values.yaml#L15-L19), so I'm not sure where to put these exclusions

adityathebe (approved):

moshloop:
@mainawycliffe 

 - [ ] I think lets remove the exclusions for now, the registry has sane defaults
 - [ ] The clusterName is not specified correctly for helm install
 - [ ] Can we move the warning inside each tab, just below, and make it warning not error style
 - [ ] Can we make kubernetes a toggle, instead of a check box
 - [ ] Switch to step based


Add these hints:

Agents are installed inside your cluster and collect data through health checks, topologies and catalog scrapers and then push the data to Mission Control. 


**Name:**
e.g. Cluster name, or if running a cluster per region/dc/account it can be the region name
**Labels**
Labels that are automatically added to all resources created by this agent
*Kubernetes* (With an icon)
Scrapes built-in and custom resource definitions and creates catalog items and a topology from  Cluster --> Namespace -->  Pod
**Interval**
How often to perform a full reconciliation of changes (in addition to real-time changes from Kubernetes events), set higher for larger clusters.


add section just after installation instructions:

See the [Kubernetes](https://docs.flanksource.com/registry/kubernetes) for more options on fine-tuning the import of Kubernetes resources. 

**Next Steps**

1. Review the dashboard and catalog pages to verify the data being imported. 
2. Create new [health checks](https://docs.flanksource.com/canary-checker/getting-started) and [catalog scrapers](https://docs.flanksource.com/config-db/overview) using Gitops in the agent cluster
3. Create new [topologies](https://docs.flanksource.com/topology/overview) to model your system -Topologies can be created either in the agent cluster or centrally to create views that span multiple agents.

PR #1670, opened 2024-02-01
1666-Canary-Page-on-Mission-Control---Health-Checks-Floating-Sidebar-Hidden-By-The-Mission-Control-Sidebar

### LINKED ISSUE #83: Improve behaviour on narrow views
Screenshot from 2021-12-08 12-33-01

When screen layout forces the selection element to the top of the screen rather than the side (eg with two windows vertically split on desktop, or on a mobile device) the selection element occupies all the vertical real estate.  It's not immediately obvious that the canary details are hidden below the fold, and a lot of space is wasted

### LINKED ISSUE #1666: Canary Page on Mission Control - Health Checks Floating Sidebar Hidden By The Mission Control Sidebar
https://www.loom.com/share/7a9fda0363b243daa7c48c0a6db70f16?sid=60af0353-d70f-4d37-bd9d-87b4660316b1

Should probably close #83 as well

mainawycliffe:
Fixes #1666
Closes #83

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,7 +79,7 @@
         "react-select": "^5.7.4",
         "react-table": "^7.7.0",
         "react-toastify": "^9.0.8",
-        "react-tooltip": "^4.5.1",
+        "react-tooltip": "^5.26.3",
         "react-top-loading-bar": "^2.3.1",
         "react-use-intercom": "^5.1.4",
         "recharts": "2.1.12",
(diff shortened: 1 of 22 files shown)


moshloop:
- We should never have a horizontal scroll, 
- The min width of the filter section is too low, even on 1280 it is very squished

moshloop:
@mainawycliffe 
- The filter popover needs an X to close it,
- Increase the breakpoint for when to show it always
- When the filter popover is used move the Stat cards to horizontal layout above the search fields,
- Split the state cards into just: Checks, Passing, Failing. - The All checks falls away and we just use filtered. 
- There is an issue with gap as well:
Screenshot 2024-02-28 at 08 26 35
- Decrease width of Tab by, Group By at lower resolutions
- Add a badge on the filter icon with the number of filters applied

moshloop:
@mainawycliffe 
- Can we use  the normal dialog close button style
- The filter badge should be number of filters enabled, not filters available
- The stat cards - take up too much Vertical height, can we make it a horizontal layout with less margins / padding and/or put them in the navbar

moshloop:
@mainawycliffe 
- [ ] The hide passing filter should also be included in the count
- [ ] Can we include the badge in the clickable target
- [ ] Can we try some design variations on the stats - its a little bland

moshloop:
@mainawycliffe 
- The count for ignore passing is inverted, it shows 1 when hide passing is unchecked
- For the counts - it should just be the total passing and total failing for the filtered view - (i.e the only time we show totals are when filters are empty)
- The X button overlaps the content, can we margin the filter body so that it never overlaps ?

PR #1684, opened 2024-02-07
fix: fix duration filter not working

### LINKED ISSUE #1691: Job History Filter Bar - Cannot unselect item / select all

mainawycliffe:
Closes #1691


moshloop:
@mainawycliffe when selecting a specific date range it doesn't format the timestamp correctly when submitting to the backlend

PR #1687, opened 2024-02-09
1682-Catalog-filters-should-carry-through-to-changesinsights-page

### LINKED ISSUE #1682: Catalog filters should carry through to changes/insights page
I.e. click on a type in the summary, see all the resources for that type and then clicking changes should show changes for that resource type only

mainawycliffe:
Closes #1682


moshloop:
@mainawycliffe - Clicking on a type does not do anything

moshloop:
@mainawycliffe - I get an `column config_changes.config_type does not exist`

PR #1698, opened 2024-02-14
testing: add login tests for saas and dashboard

mainawycliffe:
- [ ] Fix issues on the SaaS live, so we can test this propertly


e2e/user-login-flow.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { test } from "@playwright/test";
+
+const {
+  // this the base url for the kratos UI
+  PLAYWRIGHT_KRATOS_BASE_URL = "https://incident-commander.demo.aws.flanksource.com/",
```
moshloop, line 5: The URL, username and password should come from environment variables

package.json
```diff
@@ -145,6 +145,7 @@
     ]
   },
   "devDependencies": {
+    "@playwright/test": "^1.41.2",
```
moshloop, line 148: this should be in a standalone package under e2e

PR #1722, opened 2024-02-20
feat: use related_configs for displaying related configs

### LINKED ISSUE #1717: Switch to related_configs|changes functions
Depends on https://github.com/flanksource/duty/issues/445

mainawycliffe:
Fixes #1717

### DIFF
--- a/src/api/axios.ts
+++ b/src/api/axios.ts
@@ -59,6 +59,15 @@ export const Config = axios.create({
   }
 });
 
+export const Catalog = axios.create({
+  baseURL: `${API_BASE}/catalog`,
+  headers: {
+    Accept: "application/json",
+    Prefer: "return=representation",
+    "Content-Type": "application/json"
+  }
+});
+
 export const Snapshot = axios.create({
   baseURL: `${API_BASE}/snapshot`,
   headers: {
--- a/src/api/services/configs.ts
+++ b/src/api/services/configs.ts
@@ -148,6 +148,47 @@ export const getConfigChanges = (
   );
 };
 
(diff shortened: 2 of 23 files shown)


src/pages/config/ConfigList.tsx
```diff
@@ -111,7 +111,7 @@ export function ConfigListPage() {
               ...(configType
                 ? [
                     <BreadcrumbChild
-                      link={`/catalog?type=${configType}`}
+                      link={`/catalog?configType=${configType}`}
```
moshloop, line 114: Why are we changing this ?

mainawycliffe: @moshloop  This is broken, they are supposed to be configType, renamed a while back because of the conflict with change type and other type as well, so config type is clearer.

mainawycliffe:

moshloop:
@mainawycliffe For changes the relationship type can be None (Default), Incoming, Outgoing, All, when not None we need to show the config the change is for

We should also default the change duration to 7d

The relationships should also be using the recursive option

moshloop:
- When relationship type is set, we need to show the config item in the changes list 
- The date sorting doesn't appear to work
- @adityathebe can you check on the backend functions - not sure they are working correctly I expect to see the Pods of a deployment when selecting all or outgoing, and only see the replicaset

mainawycliffe:
Blocked by -> https://github.com/flanksource/duty/pull/573

mainawycliffe:
- [ ] Fix backend response issue, returns only total count

PR #1746, opened 2024-02-27
Add Playbook Action View Params

### LINKED ISSUE #1677: playbook: improvements
1. Default params should be filled with the default values or marked as optional.
image

2. 
image

3. We should also be showing the parameters on the run page
4. I think the "Run" button on the playbook run modal should be disabled until all the required params and the required resource is selected

mainawycliffe:
Closes #1677


moshloop:
Can we move it to 

Playbook

Playbook A (params)

PR #1747, opened 2024-02-27
feat: pull templated values for playbook params from API calls

### LINKED ISSUE #1745: Playbook: make a new call to fetch templated params
Once we have the playbook id, we have to make a call at /playbook/:id/params with run params (playbook_id, config_id, component_id etc.) to fetch the actual params

mainawycliffe:
Fixes #1745

### DIFF
--- a/src/components/Playbooks/Runs/Submit/SubmitPlaybookRunForm.tsx
+++ b/src/components/Playbooks/Runs/Submit/SubmitPlaybookRunForm.tsx
@@ -41,22 +41,14 @@ export default function SubmitPlaybookRunForm({
   configId
 }: Props) {
   const initialValues: Partial<SubmitPlaybookRunFormValues> = useMemo(() => {
-    const paramsDefaultValues = playbook.parameters?.reduce((acc, param) => {
-      if (param.default !== undefined) {
-        acc[param.name] = param.default;
-        return acc;
-      }
-      return acc;
-    }, {} as Record<string, any>);
-
     return {
       id: playbook.id,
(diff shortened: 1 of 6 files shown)


mainawycliffe:
@yashmehrotra 

This is the basic implementation, and works well from my initial testing, just a couple of questions

- Does this apply to parameters (or just templated parameters)
- In terms of the response, why is it double quoted?

yashmehrotra:
> @yashmehrotra
> 
> This is the basic implementation, and works well from my initial testing, just a couple of questions
> 
>     * Does this apply to parameters (or just templated parameters)
all params unless we can check on frontend if its templated or not
>     * In terms of the response, why is it double quoted?
this is json, string response can have any stringg object

src/components/Playbooks/Runs/Submit/SubmitPlaybookRunForm.tsx
```diff
@@ -118,12 +110,15 @@ export default function SubmitPlaybookRunForm({
                     <PlaybookSelectResource playbook={playbook} />
                   )
                 )}
+                {/* 
+                  We want hide params until we have a resource selected
+                */}
```
yashmehrotra, line 115: I think updating params after resource is selected makes more sense ? 
What do you think @mainawycliffe @moshloop ?

moshloop: I think hiding does make the most sense - we might have different params that are hidden and shown based on the config

PR #1748, opened 2024-02-27
1724-CodeEditor---Spec-Schemas-not-loading

### LINKED ISSUE #1724: CodeEditor - Spec Schemas not loading
This seems to be an issue affecting both monaco-yaml and monaco - conflict - may be a downgrade of monaco-yaml will work

mainawycliffe:
Closes #1724


moshloop:
@mainawycliffe this doesn't look to be working for me

moshloop:
I think for health checks lets just show the full spec - the form still only edits that first item of the type, but nothing prevents the user from adding more / different types

PR #1760, opened 2024-03-05
fix: durations should never have a prefix

### LINKED ISSUE #1757: Durations should never have a prefix
Screenshot 2024-03-05 at 07 38 45

mainawycliffe:
Fixes #1757


src/components/Playbooks/Runs/Actions/PlaybookRunsActions.tsx
```diff
@@ -67,7 +68,12 @@ export default function PlaybookRunsActions({ data }: PlaybookRunActionsProps) {
             )}
           <VerticalDescription
             label="Duration"
-            value={<Age from={data.start_time} to={data.end_time} />}
```
moshloop, line 70: I think this also applies to the per step duration  - can you also check for other durations

PR #1792, opened 2024-03-13
fix: should always redirect to the accounts url for signin

mainawycliffe:


mainawycliffe:
Also merging this for testing

PR #1814, opened 2024-03-19
Make Improvements to Configs and Fix Bugs

mainawycliffe:
- [x] default job history to two days
- [x] configs related - defaulting to outgoing
- [x] move created to the first column in changes
- [x] Relationship Type - Toggle Upstream and Incoming (Related configs and changens)
- [x] Consistent labels for config insights


moshloop:
@mainawycliffe We are missing the config type in changes view - It should only be hidden when outgoing + incoming are both off and the config is selected - for all other cases it should be shown
- Can we reduce the size of the white in the tristate, it looks a little funny

mainawycliffe:
- [ ] Changes endpoint is returning a 500 when I pass the following request -> `http://localhost:3000/api/catalog/changes?id=33316534-3334-6538-6662-613361643938&from=now-2d&to=now&page=1&page_size=200&sort_by=-created_at&depth=1&include_deleted_configs=false`

PR #1832, opened 2024-03-21
1412-CRD-backed-notifications-shouldnt-be-editable-from-the-UI

### LINKED ISSUE #1412: CRD backed notifications shouldn't be editable from the UI

### LINKED ISSUE #1810: Notifications dropdown doesn't show the connection
This is because notifications saved with crds use the connection string whereas the dropdown options use the connection id.

### LINKED ISSUE #1818: Notification summary should show the error
The response has an error `most_commons_error` field but isn't displayed on the UI

mainawycliffe:
Closes #1412
Closes #1818
Closes #1810

### DIFF
--- a/src/components/Forms/Notifications/NotificationConfigurationForm.tsx
+++ b/src/components/Forms/Notifications/NotificationConfigurationForm.tsx
@@ -1,16 +1,8 @@
 import FormikConnectionField from "../Formik/FormikConnectionField";
-import FormikTextInput from "../Formik/FormikTextInput";
 
 export default function NotificationConfigurationForm() {
   return (
     <div className="flex flex-col space-y-2">
-      <FormikTextInput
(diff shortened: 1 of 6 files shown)


src/components/Forms/Notifications/NotificationConfigurationForm.tsx
```diff
@@ -1,20 +1,24 @@
...
-      <FormikTextInput
-        name={`custom_services.0.name`}
-        label="Name"
-        hint="A unique name to identify this notification configuration"
-        required
-      />
-
-      <FormikConnectionField
-        name={`custom_services.0.connection`}
-        label="Connection"
```
moshloop, line 16: This field should be unlabelled - and the tab name should be "Connection" - The button group probably makes more sense here as well

PR #1840, opened 2024-03-25
1780-Change-sorting-should-be-done-serverdb-side

### LINKED ISSUE #1780: Change sorting should be done server/db side
Wait for new changes api

mainawycliffe:
Fixes #1780


moshloop:
@mainawycliffe It does not appear that clicking on the date column always works, if I click on another column, and then click on date it works, but then clicking again to change sort order doesn't

PR #1845, opened 2024-03-27
feat: add config relationship graph for config relationship

### LINKED ISSUE #1750: Config relationship graph
using ReactFlow

mainawycliffe:
Fixes #1750


moshloop:
@mainawycliffe its looking good

- [ ] Edges should always target the left edge from the right edge
- [ ] Edges / Shapes should not be editable
- [ ] Need the same filters as the relationship page (in fact the graph is just a button group on the relationships tab on the far right
- [ ] Group resources with more than 3 edges in a collapsible group 
- [ ] Use the component card style to highlight config status

moshloop:
@mainawycliffe 

- [ ] The incoming relationships does seem to lay out correctly -
- [ ] Clicking on a card should be take you to the configs relationship page

PR #1850, opened 2024-04-01
feat: switch to flanksource-icons

moshloop:


src/components/Forms/Formik/FormikCompactEnvVarSource.tsx
```diff
@@ -82,7 +84,7 @@ export function FormikCompactEnvVarSource({
       name: ""
     });
     setType("Static");
-  }, []);
+  });
```
mainawycliffe, line 87: I think this is may be a typo, a useEffect without a pendency array will run on every rerender, which can cause a race condition

mainawycliffe:
```
// eslint-disable-next-line react-hooks/exhaustive-deps
}, []);
```

PR #1855, opened 2024-04-02
fix: fix changes filter not working in the config details page

mainawycliffe:


mainawycliffe:
@dabasvibhor can you confirm the changes filter in the config page is working as expected

moshloop:
@mainawycliffe - Can we switch the main page to use the changes api - catalog id is no longer a required field

mainawycliffe:
- [ ] Changes API with catalog id being optional isn't deployed yet

PR #1856, opened 2024-04-03
Fix-dropdown-empty-resource-selector

mainawycliffe:


adityathebe:
@mainawycliffe the resource selector works. However there's also an invalid request to `/search` immediately after. It's sending `type: string` whereas it should be sending `types: []string`.

This is on playbook parameter type "config"

image

mainawycliffe:
@adityathebe the incorrect search would only from an incorrect spec, we pass the content of configs/checks/component directly to the resource selector, without modifying it.

image

adityathebe (approved):

mainawycliffe:
@adityathebe This could also be the reason for the second erroneous one

image

adityathebe:
> @adityathebe This could also be the reason for the second erroneous one
> 
> image

yea, I've fixed the property for the `config` parameter and it's working now.

PR #1858, opened 2024-04-04
Grouping by Multiple Rows, Dropdown Tags and Add Label and Status Column

mainawycliffe:
- [x] Enable grouping by multiple rows
- [x] Add Tags Groupby Dropdown


src/components/Configs/ConfigList/ConfigListColumn.tsx
```diff
@@ -107,20 +118,28 @@ export const configListColumns: ColumnDef<ConfigItem, any>[] = [
...
-    accessorKey: "tags",
-    cell: React.memo(ConfigListTagsCell),
-    aggregatedCell: "",
-    size: 240
-  },
-  {
-    header: "All Tags",
-    accessorKey: "allTags",
+    header: "Labels",
+    accessorKey: "labels",
```
moshloop, line 139: We should be returning or showing labels, only tags , and they should be unlabelled, if the tag is used for grouping, it shouldn't be shown either

moshloop:
- [ ] The dropdown font is not consistent - Can we not make this the same by default ?
- [ ] There should be a a group by type option, 
- [ ] The grouping should be in a single column, e.g. https://codesandbox.io/p/sandbox/react-table-grouping-by-column-ipeoh?file=%2Fsrc%2FApp.js 
Screenshot 2024-04-14 at 17 02 59

moshloop:
@mainawycliffe The relationships table now has a rendering issue

PR #1867, opened 2024-04-17
1848-Connection-Testing

### LINKED ISSUE #1848: Connection Testing
Each connection can be tested to see if it's valid or not.
Use the `/connection/test/<id>` endpoint.

Returns status codes
```
- 200 for valid
- 501 for not implemented for this particular connection type
- 500 for internal server error
```

mainawycliffe:
Fixes #1848


moshloop:
@mainawycliffe - If the connection test fails it doesn't show an error message - This should be in a popup, rather than a toast - Also move the button to the right, just before save

PR #1875, opened 2024-04-19
672-add-config-server-side-grouping-api-endpoint

mainawycliffe:


moshloop:
@mainawycliffe 
- you need to set the analysis.since, changes.since, cost:30d, 
- Type is missing from group by
- The tree should still be used
- A health summary should be added  - 


See https://github.com/flanksource/duty/issues/685

moshloop:
@mainawycliffe  some of the health seems to be duplicated

Screenshot 2024-04-26 at 12 47 42

moshloop:
It is also not clear that the types are filtered by tags

moshloop:
@mainawycliffe 

- Clicking back after selecting a grouping row, does nothing
- When grouping by more than type, there is no indicator that you filtered the data set - e.g. the tags should be preselected in the labels
Screenshot 2024-04-28 at 16 36 37
- When grouping by more than type, selecting a row, and the clicking on catalog on the breadcrumbs it takes you back to  by default sorting by type but the group by dropdown says none

moshloop:
@mainawycliffe please resolve conflicts

PR #1881, opened 2024-04-23
refactor: use API properties endpoint instead of DB properties

### LINKED ISSUE #1880: Use the /properties endpoint instread of /db/properties

### LINKED ISSUE #1885: feature flag page sends large number of requests indefinitely
[Uploading](
https://github.com/flanksource/flanksource-ui/assets/13419481/651624b5-a487-47d6-a620-41bcdfdabbc9
)

mainawycliffe:
Fixes #1880
Fixes #1885

### DIFF
--- a/src/api/query-hooks/useFeatureFlags.tsx
+++ b/src/api/query-hooks/useFeatureFlags.tsx
@@ -1,0 +1,121 @@
+import {
+  toastError,
+  toastSuccess
+} from "@flanksource-ui/components/Toast/toast";
+import { useUser } from "@flanksource-ui/context";
+import {
+  FeatureFlag,
+  PropertyDBObject
+} from "@flanksource-ui/services/permissions/permissionsService";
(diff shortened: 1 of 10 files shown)


moshloop:
@mainawycliffe - this is the wrong way around, the `/properties` should be used for feature flags, and for rendering the table, when creating / updating / deleting properties then we hit the DB endpoint - Unless `source=local` in which case you cannot create/update/delete the property as its overridden in local files on the server

mainawycliffe:
@moshloop it's currently setup like that, apart from the source, field, which is missing, should I add that on duty?

mainawycliffe:
@adityathebe can you can confirm this also fixes the following issue #1885

moshloop:
@mainawycliffe clicking on add doesn't do anything - You also should not be able to add or edit a db property if there is a local

adityathebe:
> @adityathebe can you can confirm this also fixes the following issue #1885

seems to be fixed 👍

PR #1889, opened 2024-04-29
fix: revert lifting dagre instance outside hook

mainawycliffe:
Caused unpredictable issues around the graph


mainawycliffe:
@moshloop I have hidden the nodes during the transition period to hide the weird behavior and only show the nodes once it position, what do you think

moshloop:
@mainawycliffe - If I look at https://deploy-preview-1889--goofy-euclid-75956c.netlify.app/catalog/e17e38f6-9a97-4c10-ae9f-459bd2f49792/relationships?outgoing=true it only shows the namespace

mainawycliffe:
@moshloop the intermediary node for the current config wasn't being created, hence it appeared as if it only had one node, but would expand on click

moshloop:
@mainawycliffe https://deploy-preview-1889--goofy-euclid-75956c.netlify.app/catalog/60023763-40ea-498d-ad95-6cf4c8c31e31/relationships?outgoing=true&incoming=false is only a single card?

mainawycliffe:
@moshloop I am working on fixed for those at the moment

src/components/Configs/Changes/ConfigsRelatedChanges/FilterBar/ConfigRelationshipToggles.tsx
```diff
@@ -28,6 +31,14 @@ export function ConfigRelationshipToggles() {
           setParams(params);
         }}
       />
+      <Toggle
+        label="Include Soft Relations"
```
moshloop, line 35:
```suggestion
        label="Soft"
```

src/ui/Graphs/RelationshipGraph.tsx
```diff
@@ -71,24 +72,25 @@ export function RelationshipGraph<T extends GraphDataGenericConstraint>({
...
       if (node.data.expandable) {
-        const { x, y, width, height } = getNodesBounds([node]);
-
-        // Calculate the center of the node
-        const nodeCenterX = x + width / 2;
-        const nodeCenterY = y + height / 2;
-
-        // Move the viewport to the center of the node
-        setCenter(nodeCenterX, nodeCenterY, { duration: 800, zoom: getZoom() });
+        fitView({
```
moshloop, line 86: We don't want to fit view - Node sizes should only be changed by the user, it should center on the node

PR #1890, opened 2024-04-29
fix: use `health` instead of `status` in config details

### LINKED ISSUE #1883: Use `health` instead of `status` in config details
An unhealthy config is shown as green.

image

We have added a new `health` column with possible values
- healthy
- unhealthy
- warn
- unknown
- and possibly null

Also we might need to have a display for `ready` as well @moshloop

mainawycliffe:
Fixes #1883

### DIFF
--- a/src/api/query-hooks/index.ts
+++ b/src/api/query-hooks/index.ts
@@ -171,7 +171,7 @@ function prepareConfigListQuery({
   includeAgents = false
 }: ConfigListFilterQueryOptions) {
   let query =
-    "select=id,type,config_class,status,labels,name,tags,created_at,updated_at,deleted_at,cost_per_minute,cost_total_1d,cost_total_7d,cost_total_30d,changes,analysis";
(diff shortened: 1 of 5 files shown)


adityathebe:
@moshloop do we want to show the readiness as well?

adityathebe:
@mainawycliffe doesn't look like we're using `health` in the main table?

image
image

mainawycliffe:
@adityathebe the health field is missing from the configs view

adityathebe:
> @adityathebe the health field is missing from the configs view

added it

mainawycliffe:
@adityathebe Thanks, should be good now

moshloop:
@mainawycliffe The `health` is for the dot, while the `status` is for the text and `description` is for the text below

moshloop:
@mainawycliffe - I am expected this to be "Failed", not unhealthy - This applies to the table as well
Screenshot 2024-04-28 at 16 36 37

https://deploy-preview-1890--goofy-euclid-75956c.netlify.app/catalog/822d8e05-c7c4-4822-b3c3-1a3e93b8e4ce

moshloop:
> @mainawycliffe doesn't look like we're using `health` in the main table?
> 
> image image

In that case the dot should be red, and the status should remain PodCrashloopbackoff

PR #1899, opened 2024-05-04
feat: enable auto login when password and username are provided

### LINKED ISSUE #1896: Auto login
i.e. /login?username=admin&password=test 

Kratos only

mainawycliffe:
Fixes #1896


src/components/Authentication/Kratos/KratosLogin.tsx
```diff
@@ -24,6 +24,12 @@ const Login: NextPage = () => {
   const returnToFromQuery = (query.return_to as string) || "";
   const flowId = String(query.flow || "");
 
+  const { username, password } = query;
```
moshloop, line 27: Shouldn't this be: 
```suggestion
  const { username, password, return_to } = query;
```

moshloop:
@mainawycliffe - can you check why the return_to is not working ?

PR #1902, opened 2024-05-06
fix: fix config tag display to only show value

### LINKED ISSUE #1897: Fix config tag vs label rendering
Sidebar:
- The tags get shown after the name without the label
- The labels are not shown at all 

Table:
- Tags should be shown without the label, and without show more

mainawycliffe:
Fixes #1897

### DIFF
--- a/src/components/Configs/ConfigList/Cells/ConfigListTagsCell.tsx
+++ b/src/components/Configs/ConfigList/Cells/ConfigListTagsCell.tsx
@@ -1,8 +1,6 @@
 import { CellContext } from "@tanstack/react-table";
 import { useSearchParams } from "react-router-dom";
 import { ConfigItem } from "../../../../api/types/configs";
-import Popover from "../../../../ui/Popover/Popover";
-import { TagItem, TagList } from "../../../TagList/TagList";
 
 export default function ConfigListTagsCell({
   getValue,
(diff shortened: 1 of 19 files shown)


src/components/Configs/ConfigList/Cells/ConfigListTagsCell.tsx
```diff
@@ -43,39 +41,19 @@ export default function ConfigListTagsCell({
...
-        <div className="flex flex-col items-stretch max-h-64 overflow-y-auto">
-          <TagList
-            className="flex flex-col flex-1"
-            tags={tags}
-            minimumItemsToShow={tags.length}
-          />
+    <div className="flex flex-wrap gap-1">
+      {tags.map((tag) => (
+        <div
+          className="flex flex-row p-[0.25rem] rounded-md bg-gray-200 text-gray-600 font-semibold text-xs whitespace-nowrap break-inside-avoid-column"
```
moshloop, line 51:
```suggestion
          className="flex flex-row px-1 py-0.5 rounded-md bg-gray-100 text-gray-600 text-xs whitespace-nowrap break-inside-avoid-column"
```

src/components/Configs/ConfigList/Cells/ConfigListTagsCell.tsx
```diff
@@ -43,39 +41,19 @@ export default function ConfigListTagsCell({
...
-        <div className="flex flex-col items-stretch max-h-64 overflow-y-auto">
-          <TagList
-            className="flex flex-col flex-1"
-            tags={tags}
-            minimumItemsToShow={tags.length}
-          />
+    <div className="flex flex-wrap gap-1">
+      {tags.map((tag) => (
+        <div
+          className="flex flex-row px-1 py-0.5 rounded-md bg-gray-100 text-gray-600 text-xs whitespace-nowrap break-inside-avoid-column"
```
moshloop, line 51: Can we extract this as a tag-gray class ?

moshloop: Can we use a Tags component and also add a title for the tag name

src/components/Configs/ConfigList/ConfigListColumn.tsx
```diff
@@ -90,6 +91,13 @@ export const configListColumns: ColumnDef<ConfigItem, any>[] = [
       return <Status status={health} statusText={status} />;
     }
   },
+  {
```
moshloop, line 94: I think changes should go come before tags - Can we not use  min, size and maxSizes ? i.e. don't let name / type or status too short and let the other columns shrink to shrink

PR #1911, opened 2024-05-10
fix: fix missing status for health table

mainawycliffe:


src/components/Status/index.tsx
```diff
@@ -42,7 +42,9 @@ export function Status({
   className = "",
   hideText = false
 }: StatusProps) {
-  if (!status) {
+  console.log("status", status, good, mixed);
```
moshloop, line 45:
```suggestion
```

PR #1925, opened 2024-05-15
1924-in-job-history-reduce-whitespace-around-timestamp-and-add-a-copy-button-infront-of-resource_id

### LINKED ISSUE #1924: In job history, reduce whitespace around timestamp, and add a copy button infront of resource_id

mainawycliffe:
Closes #1924


moshloop:
@mainawycliffe - I don't think we need the tooltip, can the copy button be moved to after ?

PR #1931, opened 2024-05-16
1774-Config-changes---filter-by-cell

### LINKED ISSUE #1774: Config changes - filter by cell
On hover of a  config type, config, change type, summary, created_by , add tiny `+` and `-` to update the filters

For the configs we need to provide a way to remove the filter, but we don't want an unconstrained dropdown that lists all config items (so maybe dynamically insert the items into the dropdown list)

mainawycliffe:
Fixes #1774


moshloop:
@mainawycliffe 
- Can we use the same colors for the tags in tri-state, also use a gray x instead of a trash button to remove
- When changing filters, the paging should reset to first page

src/ui/Badge/ClosableBadge.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <svg
+          className={`${svgClassName} text-blue-400" fill="${dot}" viewBox="0 0 8 8"`}
+        >
+          <circle cx={4} cy={4} r={3} />
+        </svg>
+      )}
+      {children}
+      <Button
+        icon={<IoMdClose size={12} />}
+        className="ml-1 hover:bg-gray-200 text-red-500 border-l border-gray-50"
```
moshloop, line 55: Can we update this to: hover:text-gray-700 text-gray-500 border-l border-gray-50 pl-1.5 py-1.5 text-xs rounded  (not sure where the px-2.5 comes from)

moshloop:
@mainawycliffe can we url-encode the values - Filter by summary with a `,` doesn't work

PR #1950, opened 2024-05-23
fix: reload window when the org changes

mainawycliffe:


src/components/Authentication/Clerk/ClerkAuthContextProvider.tsx
```diff
@@ -53,7 +53,12 @@ export default function ClerkAuthContextProvider({
       if (organization.id !== orgID) {
         console.log(`Clearing cache... ${orgID} -> ${organization.id}`);
         client.removeQueries();
-        setOrgID(organization.id);
+        client.refetchQueries();
+        client.resetQueries();
+        // reload the page to make sure the new organization is used
+        if (window) {
+          window.location.reload();
+        }
```
moshloop, line 61: Can't we just use client.clear()

PR #1960, opened 2024-05-28
fix: show agent details as a badge on the tables

### LINKED ISSUE #1904: We should be indicating that some of the config scrapers are from other agents.
image

mainawycliffe:
Fixes #1904


moshloop:
This badge also needs to be shown in scrapers, health checks,

PR #2012, opened 2024-06-06
fix: fix config changes filter by cell issues around encoding and colons

### LINKED ISSUE #1970: Changes Endpoint Doesn't Support URL Encoded Values For Include/Decline Data such as Summary
Here is a curl request example, with the summary field being URL encoded as it's value look like this -> `status.artifact, status.conditions`, for simple string, it works well.

Returns empty response.

```
http://localhost:3000/api/catalog/changes?summary=status.artifact%252C%2520status.conditions&from=now-2d&to=now&page=1&page_size=200&sort_by=-created_at&recursive=all&depth=1&include_deleted_configs=false
```

### LINKED ISSUE #2017: Changes - cannot remove created_by filter

mainawycliffe:
Fixes #1970
Fixes #2017

### DIFF
--- a/src/api/query-hooks/useAgentsQuery.tsx
+++ b/src/api/query-hooks/useAgentsQuery.tsx
@@ -1,16 +1,28 @@
 import { AgentSummary } from "@flanksource-ui/components/Agents/AgentPage";
 import { UseQueryOptions, useQuery } from "@tanstack/react-query";
+import { useSearchParams } from "react-router-dom";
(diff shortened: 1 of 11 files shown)


moshloop:
@mainawycliffe I think the values need to be urlencoded, then combined with a `,` and url encoded again
@adityathebe I think we todo the reverse, i.e. url decode the split match items

mainawycliffe:
@moshloop @adityathebe what about instead of a GET request, we use post and body for the filters for this endpoint? This would eliminate the need for double encoding.

PR #2015, opened 2024-06-10
1894-flanksource.ui.snippets-property

### LINKED ISSUE #1894: flanksource.ui.snippets property
This is a local only property e.g. ignore db values and don't allow creation or insertion. 

Evaluate into a function with this signature: `function(user, org)`  and

mainawycliffe:
Closes #1894


src/hooks/useFlanksourceUISnippet.tsx
```diff
@@ -0,0 +1,38 @@
...
+    if (snippets && user) {
+      try {
+        // We need to wrap the snippet in a function to be able to pass in named
+        // parameters to the snippet.
+        const wrapFn = () => `{ return  ${snippets.value} };`;
+        // eslint-disable-next-line no-new-func
+        const func = new Function(wrapFn());
+        // The first call to the function returns the actual snippet function. And
+        // the second call to the function executes the snippet function.s
+        func.call(null).call(null, user, organization);
```
moshloop, line 29:
```suggestion
        func.call(null).call(null, {user, organization});
```

src/hooks/useFlanksourceUISnippet.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [isSnippetExecuted, setIsSnippetExecuted] = useState(false);
+
+  const { featureFlags } = useFeatureFlagsContext();
+
+  useEffect(() => {
+    if (isSnippetExecuted) {
+      return;
+    }
+    const snippets = featureFlags?.find((flag) => {
+      return flag.name === "flanksource.ui.snippets";
```
moshloop, line 17: if flag.source != local, then log a warning: "UI snippets can only be set at start time. "

mainawycliffe: @moshloop at them moment, because I had added the snippet via DB, for testing purpose, it means it won't work after I add this.

PR #2034, opened 2024-06-18
Formik modal

moshloop:
Goals:

- set size, open, title, help, show*, on*, initialValues from inside the modal children
- get the modal inner body height for sizing of code editors i.e. insde the child, get height of other form fields with a ref on their wrapping div, then set height on code editor to fill reminainig
- buttons should always be pinned to footer, and have consistent styling


mainawycliffe:
@moshloop initialValues can only be set once in formik, before initialization, not after, when the form is loaded, you would need to update values, which you can do with the FormikContext, which you can setFieldValues, with the fields you need to update. I think we need to try and not deviate from this standard usage, which is already established, as long as the child component is inside a Formik form, it has access to the FormikContext and can be used to manipulate form values.

src/components/Integrations/Add/AddIntegrationForm.tsx
```diff
@@ -1,12 +1,13 @@
...
 import MissionControlRegistryOptions from "./steps/MissionControlRegistryOptions";
+import { SpecEditorProps } from "@flanksource-ui/components/SpecEditor/SpecEditor";
+import { useMemo } from "react";
 
-type Props = {
+type Props = Pick<SpecEditorProps, | "onBack"> & {
   onSuccess: () => void;
-  selectedOption: CreateIntegrationOption;
+  selectedOption: IntegrationOption;
   onBack: () => void;
```
mainawycliffe, line 11: Probably need to remove this onBack duplicate

src/ui/Modal/FormikModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { Fragment, createContext, useContext, useEffect, useRef, useState } from "react";
+import { Dialog, Transition } from "@headlessui/react";
+import { XIcon } from "@heroicons/react/solid";
+import clsx from "clsx";
+import { BsArrowsFullscreen, BsFullscreenExit } from "react-icons/bs";
+import DialogButton from "../Buttons/DialogButton";
+import HelpLink from "../Buttons/HelpLink";
+
+interface FormikModalProps extends IModalProps {
+  initialValues?: Record<string, any>;
```
mainawycliffe, line 13: This in context will be ineffective, as initial value an only be set on initial render, and the children will have access to all formik controls, to update the values, so they can update the values directly instead, if they need to.

PR #2040, opened 2024-06-20
2031-In-health-page---on-first-load-we-hide-passing-ones-by-default-However-the-Hide-Passing-toggle-is-switched-off

### LINKED ISSUE #2031: In health page - on first load, we hide passing ones by default. However the "Hide Passing" toggle is switched off.

mainawycliffe:
Fixes #2031

### DIFF
--- a/src/components/Canary/CanaryFilters/DropdownWrapper.tsx
+++ b/src/components/Canary/CanaryFilters/DropdownWrapper.tsx
@@ -1,11 +1,11 @@
 import { useAtom } from "jotai";
 import { useEffect } from "react";
-import { healthSettingsAtom } from "..";
 import {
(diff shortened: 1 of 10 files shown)


adityathebe (approved):

moshloop:
@mainawycliffe this doesn't seem to work - The UI is taking it value from query params and the query is using local storage 

Can we not just use atomWithStorage, I don't think hidePassing should update the URL state

mainawycliffe:
@moshloop we can, but I thought when sharing a link, you would want to share exactly what you looking at in terms of state?

moshloop:
@mainawycliffe - This doesn't seem to work - the toggle reflects the local storage state but the view does not

PR #2096, opened 2024-07-17
1419-direct-api-calls-to-the-backend-directly-from-the-browser

### LINKED ISSUE #1419: Direct API calls to the backend directly from the browser

mainawycliffe:
- [x] Test Saas Version

Fixes #1419


moshloop:
@mainawycliffe - This is a massive change, that is going to be very difficult to test.

Why not just make the clients non-constants and expose a api.SetBaseUrl() method that set sets `axios.defaults.baseURL`

I also don't see why we have so many - we just need a single client, adding a `/db` or `/catalog` prefix to urls would make it easier to understand what  the final path would be .

src/hooks/useClerkAttachAuthInterceptorsToAxios.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const interceptorsResponseCleanups: number[] = [];
+
+    allAxiosInstances.forEach((axiosInstance) => {
+      if (!backendUrl) {
+        return;
+      }
+
+      // register axios interceptors here
+      const requestInterceptor = axiosInstance.interceptors.request.use(
+        async (config) => {
```
moshloop, line 46: Have we confirmed that this only happens once, and not on every backend request? What happens when the token expires or is it a long lived token ?

moshloop: I created a `Backend` JWT template with a 30 day expiry that we should use, but we still shouldn't be requesting it every time.

mainawycliffe: @moshloop the getAuth method should cache the token within the token TTL, you would need to skipCache explicitly. In terms of renewal, clerk should handle that automatically for us, and we will always get an update token when we request, either from cache or a fresh if the TTL is expired.

moshloop:
@mainawycliffe please resolve conflicts

PR #2125, opened 2024-07-25
fix: add tooltip to show full tag name for tags

### LINKED ISSUE #2124: Use tooltip to show full tag name for tags
Currently we display tags as only values, if we hover over it, we should get `key:value`

mainawycliffe:
Fixes #2124

### DIFF
--- a/src/ui/Tags/Tag.tsx
+++ b/src/ui/Tags/Tag.tsx
@@ -27,40 +27,51 @@ export function Tag({
   className = "flex flex-row px-1 py-0.5 rounded-md text-xs whitespace-nowrap break-inside-avoid-column",
   variant = "gray",
   tag,
+  id,
   onFilterByTag,
   ...props
 }: TagProps) {
   return (
     <>
       <div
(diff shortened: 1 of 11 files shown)


moshloop:
@mainawycliffe Can we move the include / exclude icons into the tooltip ?

src/ui/Tags/Tag.tsx
```diff
@@ -26,35 +27,50 @@ export function Tag({
...
   return (
-    <div
-      className={clsx(
-        className,
-        variant === "gray" && "bg-gray-100 text-gray-600",
-        variant === "blue" && "bg-blue-100 text-blue-800",
-        onFilterByTag && "group"
+    <>
+      <div
+        data-tooltip-id={`${id}-${tag?.key}-${tag?.value}`}
```
moshloop, line 37:
```suggestion
        data-tooltip-id={`${id}-${tag?.key}-${tag?.value}`}
        data-tooltip-delay-show={100}
```

moshloop:
@mainawycliffe can we put the icons on their own row, middle aligned so that the mouse cursor goes up, with no chance of the tooltip disappearing before reaching the icon

PR #2126, opened 2024-07-25
Revert "1419-direct-api-calls-to-the-backend-directly-from-the-browser"

moshloop:
Reverts flanksource/flanksource-ui#2096


mainawycliffe (approved):
I managed to fix the url issue, but seems to be having an issue with postgrest and how it parses the response, as this is failing, let's revert for now and I can do a thorough testing as soon as we have a Saas testing backend

mainawycliffe:
@moshloop don't merge it, seems like a CORS issue, let me see if we can fix this one quickly

mainawycliffe:
Closing this, as we already resolved the issue

PR #2144, opened 2024-07-30
fix: improve the search mechanism for the playbook resource selector

### LINKED ISSUE #2142: Playbook dropdown search is broken, results should come in order of keyword matching priority

mainawycliffe:
Fixes #2142

### DIFF
--- a/src/api/types/playbooks.ts
+++ b/src/api/types/playbooks.ts
@@ -72,6 +72,7 @@ export type PlaybookResourceSelector = {
   statuses?: string[];
   labelSelector?: string;
   fieldSelector?: string;
+  search?: string;
 };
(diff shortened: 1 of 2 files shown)


moshloop:
I don't think this is what we want - we still want to search across those other fields - I think we should switch to remote search when filtering i.e.

```yaml
configs:
  - search: "kub" # whatever the user has entered into the box
     types:
       - Kubernetes::Deployment
       - Kubernetes::StatefulSet
       - Kubernetes::DaemonSet
```

wdyt @yashmehrotra @adityathebe

moshloop:
@mainawycliffe the search api no supports a `search` field

PR #2146, opened 2024-07-31
1738-Playbook-Option-to-re-run-a-playbook-run-with-the-same-parameters

### LINKED ISSUE #1738: Playbook: Option to re-run a playbook run with the same parameters

mainawycliffe:
Fixes #1738


moshloop:
@mainawycliffe can we change it so that after (re)-running a playbook it takes you directly to the page instead of the toast

PR #2147, opened 2024-07-31
Fix-resource-tables-issues

mainawycliffe:


moshloop:
@mainawycliffe the name column is too short - Can we also make all columns resizable again, with a max width on the date columns

PR #2175, opened 2024-08-08
2169-Tags---Should-be-able-to-include-a-cluster-tag-but-then-exclude-1-or-more-namespace-tags

### LINKED ISSUE #2169: Tags - Should be able to include a cluster tag, but then exclude 1 or more namespace tags

mainawycliffe:
Fixes #2169

### DIFF
--- a/src/api/query-hooks/index.ts
+++ b/src/api/query-hooks/index.ts
@@ -365,7 +365,7 @@ export function useGetConfigTagsListQuery(
 
 export function useGetConfigLabelsListQuery() {
   return useQuery({
-    queryKey: ["configs", "tags", "list"],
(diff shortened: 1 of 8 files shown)


moshloop:
@mainawycliffe this doesnt add the `!` when submitting to the server

mainawycliffe:
> @mainawycliffe this doesnt add the `!` when submitting to the server

Sorry Moshe, forgot to mark as draft, still testing this and for some reason, the performance issues are worse with devtool open on my side, let me sort that out and then test the PR.

moshloop:
I can still include and exclude in the same tag
Screenshot 2024-08-08 at 20 54 43

PR #2185, opened 2024-08-12
fix: don't show crd link, when source is Push

### LINKED ISSUE #2184: In settings > topology, If source is Push CRD link should not be shown (it crashes) and allow option to delete
Example in last line: https://app.flanksource.com/settings/topologies/204c57e8-2226-4f2b-a53d-61c88f7ab580

mainawycliffe:
Fixes #2184

### DIFF
--- a/src/components/Settings/CRDSource.tsx
+++ b/src/components/Settings/CRDSource.tsx
@@ -16,7 +16,7 @@ export default function CRDSource({
   source,
   showMinimal = false
 }: CRDSourceProps) {
-  if (!source || source === "UI") {
+  if (source?.toLowerCase() !== "KubernetesCRD".toLowerCase()) {
     return null;
   }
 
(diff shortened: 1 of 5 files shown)


src/components/Settings/CRDSource.tsx
```diff
@@ -16,7 +16,7 @@ export default function CRDSource({
   source,
   showMinimal = false
 }: CRDSourceProps) {
-  if (!source || source === "UI") {
+  if (!source || source === "UI" || source.toLowerCase() === "push") {
```
yashmehrotra, line 19: I think we should inverse this logic, ie

```
if source != "crd" {return null}

return (
  <comp>
)
```

moshloop:
@mainawycliffe  There is no save/update options for UI topologies now ?

moshloop:
Can we change the link for agent to be `RiRobot2Fill Agent`

moshloop:
@mainawycliffe  - It is wrapping, has an underline and it says `link` instead of `Agent`

PR #2216, opened 2024-08-23
2174-Merge-topology-and-catalog-pages

### LINKED ISSUE #2174: Merge topology and catalog pages
- If there is only a single catalog for a topology show the merged view

mainawycliffe:
- [ ] Add and switch to using the components inside config details
- [ ] Improve the wrapping of the topology properties and analysis 

Fixes #2174


moshloop:
@mainawycliffe Can we extract the topology card changes to their own card ?

moshloop:
@mainawycliffe 
- [ ]  Can we use a splitter in between? https://github.com/johnwalley/allotment 
- [ ] the headline properties should render exactly the same at the very top of the panel, 
- [ ] There is a double tooltip on the metrics, and the bar should be left aligned
Screenshot 2024-10-21 at 10 23 51

PR #2234, opened 2024-08-30
2229-Run-now---error-messages

### LINKED ISSUE #2229: Run now - error messages
- Something went wrong: ["check","details",null] data is undefined
- invalid input syntax for type uuid: "undefined"
- Remove success message

mainawycliffe:
Fixes #2229


mainawycliffe:
@moshloop About this error: `Something went wrong: ["check","details",null] data is undefined`, it occurs if the network connection breaks, we probably need to handle it globally

PR #2264, opened 2024-09-06
fix: refresh action queries, on click refresh on navbar

### LINKED ISSUE #2257: Playbook run action - incorrect query
When a playbook is running I see both a 

`api/db/playbook_run_actions?select=id,name,status,start_time,end_time,scheduled_time`

and a 
 
`/api/db/playbook_run_actions?select=id,name,status,start_time,end_time,scheduled_time&order=start_time.asc&playbook_run_id=eq.0191bce0-be08-45fa-8a4c-e8d9f77236f5`

However I don't see 
`api/db/playbook_run_actions?id=eq.0191bcc8-ccb0-af5c-e1f1-14dfeaca3042&select=*`

Which seems to be called only on first selecting an action - not on a backround or forced refresh

mainawycliffe:
Fixes #2257


moshloop:
@mainawycliffe - the caching of the details seems fixed, but the refresh does not seem to work

If I run the the delayed exec playbook with a value of "15s", I expect to see a regular query for playbook_runs (Maybe once per second when the status is not in completed/failed) and then for playbook_run_actions every couple seconds, to pick up new actions that have run 

The result should be:

Run: Scheduled or Sleeping (depending on how fast the server) that transitons to failed
Delay Action - To come up on its as failed

PR #2277, opened 2024-09-11
feat: highlight deleted configs and improve how they deleted configs are shown

### LINKED ISSUE #2249: Config Page - Highlight deleted configs
- Add a Trash Icon into breadcrump with tooltip > Deleted: timestamp
- Remove standalone deleted_at property
- Change status to:

Status
{Trash} Deleted {age}

- Rename existing status to Last Status, and place below status

mainawycliffe:
Fixes #2249


moshloop:
@mainawycliffe In the relationships table - there is no indicator for deleted items

mainawycliffe:
@moshloop you mean in the status column? I think the function isn't returning the deleted_at field back, which is used to determine that

PR #2294, opened 2024-09-17
feat: add Notification List with settings and selenced tabs

### LINKED ISSUE #2286: Notification List
- Add a notifications icon top right

Tabs: 
- Notifications (With indicators silenced) and count + firstObserved handling
   - List of notifications
   - Filters (Sent/Failed/Silenced/Repeated) 
- Rules (Moved from Settings)
- Silences (With add/delete button)

mainawycliffe:
- [x] Pull in correct data from API (will ask in stand-up

Fixes #2286


adityathebe:
The count only makes sense for unsent notifications (silenced, repeat interval)

Maybe the count should be besides the status like we have for configs.

image

moshloop:
@mainawycliffe - Can we add an icon and modal for the errors ? - The status filter also needs to be fixed

PR #2315, opened 2024-10-01
2307-When-inviting-a-user-the-assigned-role-is-not-persisted

### LINKED ISSUE #2307: When inviting a user, the assigned role is not persisted.

mainawycliffe:
Fixes #2307

### DIFF
--- a/src/pages/UsersPage.tsx
+++ b/src/pages/UsersPage.tsx
@@ -1,19 +1,14 @@
 import useUpdateUser, {
-  useDeleteUser
+  useDeleteUser,
+  useInviteUser
 } from "@flanksource-ui/api/query-hooks/useUserAPI";
 import { RegisteredUser } from "@flanksource-ui/api/types/users";
-import UserForm, {
-  UserFormValue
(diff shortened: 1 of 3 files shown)


src/pages/UsersPage.tsx
```diff
@@ -193,6 +178,45 @@ export function UsersPage() {
...
+            onClose={() => {
+              setInviteLink(undefined);
+            }}
+            open={!!inviteLink}
+            bodyClass=""
+            size={"small"}
+          >
+            <div className="flex flex-col bg-white p-4">
+              <div className="p-2 text-sm">
+                User has been invited successfully. An email has been sent to
```
moshloop, line 193: Why do we have this?  The invite link is considered a password and shouldn't be shared with anyone.

moshloop: Thinking about this more, I think we should keep it in

moshloop:
@mainawycliffe - Can we a copy button ?

mainawycliffe:
@moshloop  The copy button is there

image

PR #2325, opened 2024-10-02
fix: update UI to reflect to configs view changes

### LINKED ISSUE #2324: In `/catalog/?config_type=<config_type>` page, which calls `/api/db/configs`, the changes and analysis returned will be of the same type(schema) as catalog summary request (changes will be count of changes and analysis will be map of analysis type to count), so the UI needs to be updated to display those correctly

mainawycliffe:
Fixes #2324


mainawycliffe:
Merging this, to fix UI crashing on the catalog page

PR #2326, opened 2024-10-03
fix: update message for when the user hasn't started typing and options are empty

### LINKED ISSUE #2313: Notification Silence Resource dropdown is initially empty
- It should behave same as playbook (later on we might more inteliigence to what gets shown there) and
- the "No options" should be changed to "Enter X or more characters, and then only no options found"

image

mainawycliffe:
Fixes #2313


adityathebe:
@mainawycliffe the messages are working correctly. Although I am unable to unselect.

mainawycliffe:
@adityathebe should be better now

adityathebe:
LGTM @mainawycliffe 

cc @moshloop

PR #2328, opened 2024-10-04
fix: fix config relationship table crashing due to change in API

mainawycliffe:


mainawycliffe:
Merging this to fix an issue, and bring it into another PR -> #2329

PR #2329, opened 2024-10-04
2285-Catalog-Detail-Page---Add-pagination

### LINKED ISSUE #2285: Catalog Detail Page - Add pagination

mainawycliffe:
Fixes #2285


moshloop:
@mainawycliffe  

- Can we disable hiding and drag to group on these tables as it doesn't work, 
- for relationships table can we set max width for expand column, and reduce the padding - the rows should be same height as other tables

PR #2354, opened 2024-10-17
Fix-tables-crashing

mainawycliffe:


mainawycliffe:
Merging this as the agents, notification rules and silences tables are crashing

PR #2358, opened 2024-10-18
fix: fix sorting and paging not working on config changes

mainawycliffe:


moshloop:
@mainawycliffe it doesn't look like the page size is included in the first request

moshloop:
@mainawycliffe can we default the sort on changes to be by last seen ?

mainawycliffe:
@moshloop the backend only allows name, change_type, summary, source, created_at, count for sorting

moshloop:
@mainawycliffe - created_at is last seen - before it was sorted by most recent, and now its unsorted which doesn't make sense when looking at history

PR #2372, opened 2024-10-23
fix: fix edit Silence durations fails with date math

### LINKED ISSUE #2369: Edit Silence durations fails with date math

mainawycliffe:
Fixes #2369


moshloop:
@mainawycliffe  When I edit the duration it works, but then I click on the row again and it shows the old duration

PR #2373, opened 2024-10-23
fix: remember page sizes for tables on local storage

### LINKED ISSUE #2362: Config changes - remember page size
Currently when filtering items out, it resets the page size - But it should be using localstorage to save it

mainawycliffe:
Fixes #2362


moshloop:
@mainawycliffe If you goto the 2nd page (where it adds a pageSize header) and then filter out a value it resets to 50

PR #2374, opened 2024-10-23
fix: fix column size and padding in configs tables

### LINKED ISSUE #2363: Config relationships - padding
- Reduce the width of the collapse column
- Reduce padding around cells - cell height should be consistent between relationships / cjanges

mainawycliffe:
Fixes #2363


moshloop:
1. The cell heights / padding are not consistent
2.  There is a also a grayed our right chevron in the rows of the relationships which shouldn't be there

Screenshot 2024-10-25 at 13 37 12
Screenshot 2024-10-25 at 13 36 41
Screenshot 2024-10-25 at 13 36 21

PR #2490, opened 2025-03-19
feat: Playbook cancellation

adityathebe:
related: https://github.com/flanksource/mission-control/issues/1911


src/components/Playbooks/Runs/CancelPlaybookButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const [showCancelModal, setShowCancelModal] = useState(false);
+
+  const terminalStatuses = ["cancelled", "completed", "failed"];
+  if (terminalStatuses.includes(status)) {
+    return null;
+  }
+
+  return (
+    <>
+      <Button
```
moshloop, line 30: Should be able to use just the ConfirmationButton

moshloop: Or DeleteButton - just need to make icon configurable

PR #2499, opened 2025-03-21
feat: Show the spec in the runs page

adityathebe:
resolves: https://github.com/flanksource/flanksource-ui/issues/2497


moshloop:
@adityathebe  Can we put both Spec and Params under a 3-dot menu on the far right?

adityathebe:
image

---

image

PR #2500, opened 2025-03-21
feat: display artifacts on playbook actions

adityathebe:
related: https://github.com/flanksource/mission-control/issues/1961

https://github.com/user-attachments/assets/e7453ef7-d78d-4622-a5ac-ec005c20cb73

### DIFF
--- a/src/api/axios.ts
+++ b/src/api/axios.ts
@@ -136,6 +136,14 @@ export const PlaybookAPI = axios.create({
   }
 });
 
+export const ArtifactAPI = axios.create({
+  baseURL: `${API_BASE}/artifacts`,
+  headers: {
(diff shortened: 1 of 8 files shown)


moshloop:
I think we will also need to show something in the header that their are artifacts 
Screenshot 2025-03-21 at 17 18 20

It probably makes more sense to have the tab name be the artificact filename, with an extra tab per artifact.

For large (> 1mb) and unsupported formats (.txt, .yml, yaml, json, md, .log, .xml) we shouldn't display inline and just show a download link

adityathebe:
image

---
image
---

image

adityathebe:
Show the artifacts count in the header

image

PR #2523, opened 2025-03-28
feat: show health & status of resources in notification send history

adityathebe:
resolves: https://github.com/flanksource/flanksource-ui/issues/2522

<details>
<summary>Old</summary>

image
</details>

**New**

image


moshloop:
@adityathebe Can we rename the existing column to notification and then add a status column so that the UI is consistent ?

PR #2554, opened 2025-04-30
feat: notification summary

adityathebe:


moshloop:
@adityathebe it doesnt look like the status filters work on the dtail page

moshloop:
@adityathebe We seem to be missing the count in the last seen

PR #2561, opened 2025-05-08
fix: kratos backend url for kratos-client-fetch

adityathebe:


adityathebe:
Merging this because the euclid deployment merges 2 `Set-Cookie` headers into 1.

PR #2691, opened 2025-10-30
Add BarGauge panel type

adityathebe:
## Summary
Adds a new BarGauge panel type for displaying multiple horizontal bar gauges in a single panel, similar to Grafana's BarGauge visualization. Supports threshold-based coloring and values exceeding 100%.

image


adityathebe:
image

added support for grouping. Because we need different thresholds for real, limit and request.

Issue #2678, opened 2025-10-15
Connection: Editor form is broken


adityathebe:
<img width="1547" height="596" alt="Image" src="https://github.com/user-attachments/assets/b451fe0c-908f-447f-95e3-4d878dbad2a5" />

We have dedicated custom forms for some connection types (like HTTP, Prometheus, ...) however some forms are just generic yaml editor.

Those generic YAML editors are currently broken. They are not taking the full height of the form.
I barely see one line.

Some types with generic form: `gemini`, `azure_key_vault`, `loki`, ...

PR #2807, opened 2025-12-30
fix: add missing form. Cover all connection types

adityathebe:
resolves: https://github.com/flanksource/flanksource-ui/issues/2678

## Summary by CodeRabbit

* **New Features**
  * Permission-based controls added for Save/Update and Delete actions in the connection editor.
  * Read-only display and warning for connections sourced from Kubernetes CRDs.
  * New connection types introduced (Gemini, Loki, OpenSearch, AWS/Azure/GCP KMS and related vaults) with corresponding form support.

* **Chores**
  * Updated a UI icon dependency to a newer patch version.

* **API/Schema**
  * Connection form now accepts additional optional fields (e.g., URLs, key ID, index, NTLM/Digest/TLS flags).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@flanksource/flanksource-ui",
-      "version": "1.4.148",
+      "version": "1.4.149",
       "dependencies": {
         "@ai-sdk/anthropic": "^3.0.1",
         "@ai-sdk/mcp": "^1.0.1",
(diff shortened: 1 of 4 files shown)


moshloop:
@adityathebe When looking at the gemini connection it shows created_at, id, metadata fields shouldn't it only be spec ?
We also need a gemini connection

PR #2849, opened 2026-02-11
feat(playbooks): add AI diagnose failure button for failed playbook runs

adityathebe:
Adds a **Diagnose Failure** button to the playbook run detail page when a run has failed.

resolves: https://github.com/flanksource/mission-control/issues/2744


moshloop:
@adityathebe make the button size consistent with rerun playbook

PR #2984, opened 2026-04-14
chore: reduce notification summary latency from repeated refetches

### LINKED ISSUE #2982: Increase cache staleTime for notification summary
/summary has high latency and stale age = 0.

adityathebe:
/notification/summary is relatively slow, and immediate staleness was triggering unnecessary refetches. This change keeps results fresh for a short window so users see faster loads and fewer redundant backend calls.


moshloop:
@adityathebe, I think we should make this an incremental refetch on a short window, and full refetch every 5m, ideally this should even be realtime

PR #2987, opened 2026-04-17
feat: add jobs summary page with drilldown

### LINKED ISSUE #2963: Properties to disable job
I think it'll be useful to stop some jobs from running via properties.

Yesterday, ProcessFallbackNotifications had a bug where it continuously emitted notifications.
Having the ability to stop just one job would have been very helpful in that case.

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/2963

## Summary by CodeRabbit

* **New Features**
  * Jobs history summary page with aggregated metrics (avg duration, counts), server-side sorting and pagination.
  * Per-job overrides dialog to view/edit job settings, including enable/disable toggle.
  * Drill-down settings page and navigation to view detailed job history with pagination and refresh.
  * Filters updated to optionally hide job-name selector and accept a default status filter.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -286,6 +286,10 @@ const JobsHistorySettingsPage = dynamic(
   () => import("./components/JobsHistory/JobsHistorySettingsPage")
 );
 
+const JobsHistoryDrilldownSettingsPage = dynamic(
+  () => import("./components/JobsHistory/JobsHistoryDrilldownSettingsPage")
+);
+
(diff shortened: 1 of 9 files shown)


moshloop:
@adityathebe can we do 3 dots instead of Manage so that the rows don't become so tall ?

adityathebe:
image

keeping the button as I think that looks better. I've reduced the row height.

image

PR #3011, opened 2026-04-30
feat: add scatter-plot timeline view for config changes

adityathebe:
Another approach here: https://github.com/flanksource/flanksource-ui/pull/3013


adityathebe:
https://github.com/flanksource/flanksource-ui/pull/3013

